PR #2, opened 2018-02-27
[Fastlane] Set version and bump build number in the single target

dcordero:
Another iteration on issue #130

The internal implementations of `increment_version_number` and `increment_build_number` actions are based on agvtool which bumps the version and build number in all the targets of the project.

Using PlistBuddy on the other hand, it is possible to set keys on a specific Info.plist. 

- [x] Bump build number only the in the Info.list of the appropriate platform
- [x] Set version only in the Info.plist of the appropriate platform
 
Great idea (of course not mine), coming from Timehop and found in fastlane/examples project:
https://github.com/fastlane/examples/blob/master/Timehop/Fastfile


fastlane/helpers/VLC.rb
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/ruby
+
+class VLC
+
+  class InfoPlistPath
+    IOS = '../Sources/VLC for iOS-Info.plist'
```
carolanitz: For iOS we also need to update the watchkit and the watchkit extension with the same number :)

fastlane/Fastfile
```diff
@@ -1,40 +1,46 @@
...
-  bump_version_and_commit if platform == 'iOS'
-  git_apply_private_constants
-  cocoapods(repo_update: true)
-  gym(scheme: "VLC-#{platform}")
-  pilot(app_platform: platform == 'tvOS' ? 'appletvos' : 'ios')
-  update_changelog
-  push_to_git_remote
+   platform = get_platform options
+   version = get_version options
+
```
carolanitz: a newline slipped in :D

dcordero:
Good catch @carolanitz, thanks for reviewing. I have updated the code to also bump the missing the plist for watchkit and watchkit extension.

carolanitz:
☺️Yay LGTM! Thanks for reworking this. I'm gonna try it out Monday!

carolanitz:
this was actually already merged

PR #5, opened 2018-03-01
make VLC compile with Xcode 9.3 beta 2

carolanitz:


fkuehne:
In case the payload is nonnull, looks good to me!

carolanitz:
It was actually complaining since payload is an id<NSObject,NSCoding> that you can't use VLCWatchMessageKeyURIRepresentation as key. I'm not sure why @toco defined it that way, I'm guessing watch constraints. I didn't want to change it but we're always passing in an NSDictionary so this seemed the savest

PR #6, opened 2018-03-01
info.plist: remove the public.txt and public.plain-txt contentTypes

carolanitz:
We had complaints by a lot of users that VLC gets chosen for .txt files.
Even though we support this format (because we obviously support all the formats 😜) we shouldn't list them because in the supported content types. so this PR

- [x] removes .txt from the publicly supported file types

### DIFF
--- a/Sources/VLC for iOS-Info.plist
+++ b/Sources/VLC for iOS-Info.plist
@@ -82,7 +82,6 @@
 			<string>Owner</string>
 			<key>LSItemContentTypes</key>
 			<array>
-				<string>public.plain-text</string>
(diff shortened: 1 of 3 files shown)


carolanitz:
I've been breaking things because I don't know about all parts of the codebase still. You might have told me "there is also that file that you forgot" or "removing plain-text might maddle with opening srt files" (but I tested that), So I really appreciating you or anybody else looking over every commit from now on :) Also you and Bubu have an easier time seeing what I change + knowledge transfer

Mikanbu:
LGTM as well!

carolanitz:
merged

PR #7, opened 2018-03-01
360videos with device motion

carolanitz:
It's still work in progress but I wanted to share this. 
I'm still having the problem that when I hold the phone exactly opposite of the starting point I see the exact start frame upside down again and currently the gesturerecognizers are completely overwritten. 

[360videos with devicemotion](https://youtu.be/AKYC6pecZxk)


carolanitz:
btw here are videos for testing : https://www.mettle.com/360vr-master-series-free-360-downloads-page/

lazerwalker:
I'm really rusty at this sort of math, but that rotation problem sounds a bit like the sorts of behaviors you see when encountering [gimbal lock](https://en.wikipedia.org/wiki/Gimbal_lock). 

If that's the case, the solution would be to switch from using Euler angles to using Quaternions. The math is a bit of a pain, but they are exposed as part of the `CMAttitude` data.

carolanitz:
before merging this PR still needs to address the following: 
- [ ] work together with gesture recognizers
- [ ] have a disabled rotation lock

carolanitz:
This works now with gesture recognizer and devicemotion BUT it feels weird once you use devicemotion again after panning. I can't say why that is exactly but I feel like that could go into a follwo up commit

magwyz:
Hum, I thought my patch was working well...
Did you try to add the support from something else or my patch is broken?
I am not sure to see what is the issue on your video.

carolanitz:
@magwyz Yes your patch worked fine for the devicemotion alone. But once I add panning to the mix and change the starting point for the devicemotion it seems to experience some axe switching :(

magwyz:
@carolanitz Panning is moving the video with the fingers, right?
About the starting point, you want to add a set constant angle to the rotations because you do not like the current starting point, right?

carolanitz:
yes panning is moving with fingers and I don't want to add a set angle to the rotations. I'm just saving the location the user panned to and then add the devicemotion diff to that position. but that seems to mess up roll and yaw somehow

carolanitz:
@magwyz I think it's okay for a first version but I will create a ticket for followup work.

magwyz:
@carolanitz Ok, I understand what you want to achieve but I have no fix in mind when I read the MR. Do as you want for the first version. If you want, I will also be able to have a look when I go back to work next week and have access to the Mac and a device.

carolanitz:
@magwyz that's basically what we decided on :) it already feels so much better than what we have currently 😄

magwyz:
@carolanitz Sorry, I don't get it. What have you decided? What is better than currently?

carolanitz:
That we merge it in as is and these changes from this Pull request are better than what we currently have for 360 in the shipped Appstore app. We can then follow up with a patch release to make it perfect once you're back

magwyz:
ok, fair enough. :-)

carolanitz:
merged with : 
3d5dc239 VLCMovieViewController: 360 videos fix rotationlock
f125c56c 360 video: adjust values when they get out of bounds
7da7983f DeviceMotion: resolved jumping & not sensitive enough panning
df0d5046 DeviceMotion: make panning and devicemotion interact with each other
4a38e44d 360Video: moved currentMediaIs360 to playbackController
9563d3af DeviceMotion: use quaternions and convert them ourself
82f8e9d1 Devicemotion: add devicemotion for 360 videos

PR #8, opened 2018-03-02
VLCKeychainCoordinator: if the passcode is empty we won't show the pa…

carolanitz:
…sscode screen

We had some users that said they were prompted by the passcode screen even though they never set a passcode. I just ran into that issue. This will fix the presentation but still not the underlying bug which is that passcode lock is all of a sudden enabled in the settings but we don't have a passcode. 

I am really not sure what is causing the wrong settings value from the userdefaults tbh


carolanitz:
the pattern is after an update but that "shouldn't" affect the userdefaults

PR #9, opened 2018-03-03
[Playback] iOS control center media controls

mkchoi212:
It seems like simply setting ` _mediaPlayer.position` does not actually change the position of control center's scrubber; or am I missing something???

My workaround was using the `[VLCMediaPlayer jumpFoward:Int]` function to take care of all the magic that happens behind the scenes for me.

### DIFF
--- a/SharedSources/VLCRemoteControlService.m
+++ b/SharedSources/VLCRemoteControlService.m
@@ -27,6 +27,7 @@ static inline NSArray * RemoteCommandCenterCommandsToHandle()
              cc.skipForwardCommand,
              cc.skipBackwardCommand,
              cc.changePlaybackRateCommand,
+             cc.changePlaybackPositionCommand,
              ];
 }
 
@@ -56,7 +57,7 @@ static inline NSArray * RemoteCommandCenterCommandsToHandle()
     commandCenter.changeShuffleModeCommand.enabled = NO;
     commandCenter.seekForwardCommand.enabled = NO;
     commandCenter.seekBackwardCommand.enabled = NO;
-    commandCenter.changePlaybackPositionCommand.enabled = NO;
+    commandCenter.changePlaybackPositionCommand.enabled = YES;
 
     NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
     NSNumber *forwardSkip = [defaults valueForKey:kVLCSettingPlaybackForwardSkipLength];
@@ -123,6 +124,11 @@ static inline NSArray * RemoteCommandCenterCommandsToHandle()
     if (event.command == cc.changePlaybackRateCommand) {
         MPChangePlaybackRateCommandEvent *rateEvent = (MPChangePlaybackRateCommandEvent *)event;
         [_remoteControlServiceDelegate remoteControlService:self setPlaybackRate:rateEvent.playbackRate];
+        return MPRemoteCommandHandlerStatusSuccess;
+    }
+    if (event.command == cc.changePlaybackPositionCommand) {
+        MPChangePlaybackPositionCommandEvent *positionEvent = (MPChangePlaybackPositionCommandEvent *)event;
+        [_remoteControlServiceDelegate remoteControlService:self setCurrentPlaybackTime:positionEvent.positionTime];
         return MPRemoteCommandHandlerStatusSuccess;
     }
     NSAssert(NO, @"remote control event not handled");
--- a/SharedSources/VLCRemoteControlService.m
+++ b/SharedSources/VLCRemoteControlService.m
@@ -129,10 +129,12 @@ static inline NSArray * RemoteCommandCenterCommandsToHandle()
         [_remoteControlServiceDelegate remoteControlService:self setPlaybackRate:rateEvent.playbackRate];
         return MPRemoteCommandHandlerStatusSuccess;
     }
-    if (event.command == cc.changePlaybackPositionCommand) {
-        MPChangePlaybackPositionCommandEvent *positionEvent = (MPChangePlaybackPositionCommandEvent *)event;
-        [_remoteControlServiceDelegate remoteControlService:self setCurrentPlaybackTime:positionEvent.positionTime];
-        return MPRemoteCommandHandlerStatusSuccess;
+    if (@available(iOS 9.1, *)) {
+        if (event.command == cc.changePlaybackPositionCommand) {
+            MPChangePlaybackPositionCommandEvent *positionEvent = (MPChangePlaybackPositionCommandEvent *)event;
+            [_remoteControlServiceDelegate remoteControlService:self setCurrentPlaybackTime:positionEvent.positionTime];
+            return MPRemoteCommandHandlerStatusSuccess;
+        }
     }
     NSAssert(NO, @"remote control event not handled");
     APLog(@"%s Wasn't able to handle remote control event: %@",__PRETTY_FUNCTION__,event);
--- a/SharedSources/VLCRemoteControlService.m
+++ b/SharedSources/VLCRemoteControlService.m
@@ -32,7 +32,7 @@ static inline NSArray * RemoteCommandCenterCommandsToHandle()
     if (@available(iOS 9.1, *)) {
         [commands addObject:cc.changePlaybackPositionCommand];
     }
-    return commands;
+    return [commands copy];
 }
 
 - (void)subscribeToRemoteCommands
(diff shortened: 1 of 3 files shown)


SharedSources/VLCRemoteControlService.m
```diff
@@ -56,7 +57,7 @@ - (void)subscribeToRemoteCommands
     commandCenter.changeShuffleModeCommand.enabled = NO;
     commandCenter.seekForwardCommand.enabled = NO;
     commandCenter.seekBackwardCommand.enabled = NO;
-    commandCenter.changePlaybackPositionCommand.enabled = NO;
+    commandCenter.changePlaybackPositionCommand.enabled = YES;
```
carolanitz: I think YES is the defaultvalue and you can simply remove this line :)

Sources/VLCPlaybackController.m
```diff
@@ -1106,8 +1108,10 @@ - (void)_recoverLastPlaybackStateOfItem:(MLFile *)item
         CGFloat lastPosition = .0;
         NSInteger duration = 0;
 
-        if (item.lastPosition)
+        if (item.lastPosition) {
             lastPosition = item.lastPosition.floatValue;
+        }
```
carolanitz: 👍

Sources/VLCPlaybackController.m
```diff
@@ -504,7 +504,9 @@ - (float)playbackPosition
 
 - (void)setPlaybackPosition:(float)position
 {
-    _mediaPlayer.position = position;
+    float oldPosition = _mediaPlayer.position;
+    float diff = position - oldPosition;
+    [_mediaPlayer jumpForward:(diff * self.mediaDuration)/1000.];
```
carolanitz: we should definitely fix the setting of the mediaplayer.position here instead!

SharedSources/VLCRemoteControlService.m
```diff
@@ -27,6 +27,7 @@ @implementation VLCRemoteControlService
              cc.skipForwardCommand,
              cc.skipBackwardCommand,
              cc.changePlaybackRateCommand,
+             cc.changePlaybackPositionCommand,
```
carolanitz: this will also need an iOS  9.1 check :)

Sources/VLCPlaybackController.m
mkchoi212: Should I remove the hack or just leave a `#warning` there?

carolanitz: definitely remove it once the _mediaplayer.position works again!

mkchoi212:
@carolanitz I applied the requested fixes! Please review :)

SharedSources/VLCRemoteControlService.m
```diff
@@ -18,17 +18,21 @@ @implementation VLCRemoteControlService
...
+                                cc.nextTrackCommand,
+                                cc.previousTrackCommand,
+                                cc.skipForwardCommand,
+                                cc.skipBackwardCommand,
+                                cc.changePlaybackRateCommand,
+                                nil];
+    if (@available(iOS 9.1, *)) {
+        [commands addObject:cc.changePlaybackPositionCommand];
+    }
+    return commands;
```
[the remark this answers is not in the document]
ePirat: Wouldn't that mean the receiver has to typecast the `NSArray` to a `NSMutableArray` for this to be a problem?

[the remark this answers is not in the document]
ePirat: I've checked and Apple guidelines say:

> To determine whether it can change a received object, the receiver of a message must rely on the formal type of the return value. If it receives, for example, an array object typed as immutable, it should not attempt to mutate it.

[Source](https://developer.apple.com/library/content/documentation/General/Conceptual/CocoaEncyclopedia/ObjectMutability/ObjectMutability.html#//apple_ref/doc/uid/TP40010810-CH5-SW67)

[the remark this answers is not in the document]
carolanitz: yes they would have to typecast it and yes they shouldn't attempt to mutate it. Adding a copy is still good practice in this case since it prevents them from being able to do that. You will also find this practice in Apple's code examples :)

ePirat:
I've looked into `_mediaPlayer.position` and it works fine. Probably a feedback loop is happening here, which causes the problems.
Will try this later on an actual device (Simulator won't show the control center thing) and try to figure it out.

ePirat:
Update: I just tried it on my phone and I am not experiencing any issue at all when using the control center controls with this patch but modified to use `_mediaPlayer.position = position;`

mkchoi212:
@ePirat Setting `_mediaPlayer.position` to a new float value doesn't do anything for me. `_mediaPlayer.position` remains as `0`.

ePirat:
Works fine here, after all this is exactly how all of the scrubbing in the player currently works.

ePirat:
What files have you tried it with?

mkchoi212:
Mainly https://github.com/videolan/vlc-ios/blob/bc423a86db68cb255df0fcc95f358bd7558f035d/Sources/VLCPlaybackController.m#L505

When you resume the song after having closed it, simply setting `_mediaPlayer.position = position` within `setPlaybackPosition` does not reflect itself on the Control Center :(

I have also confirmed that the incoming value `position` is a valid float value

ePirat:
Sorry but I do not really understand what you mean. I just tried this exact method and it behaves completely correct here and sets the position as expected.

Can you please clarify the

> When you resume the song after having closed it

part?

mkchoi212:
@ePirat Here's a video of what I mean https://vimeo.com/259246384.

Notice how the position of the slider in Control Center is not correct on initial launch. It snaps into place only when I press the pause button.

ePirat:
And yes, when you set `_mediaPlayer.position = position`, the `_mediaPlayer.position` can still be a different value afterwards, given how the seeking works, it is not immediate.
So this API is not rally a great choice, IMHO, as it gives the impression it would be immediate.

ePirat:
> Notice how the position of the slider in Control Center is not correct on initial launch. It snaps into place only when I press the pause button.

That seems to be an issue with it not getting updated correctly, the code in the iOS app is really confusing when it comes to that stuff… I still do not really figured out where it actually _updates_ the position for the Control Center…

mkchoi212:
@ePirat Could you point me to a right file? I will go ahead and try to debug it!

ePirat:
I have no clue where the code for this could be… But it's definitely something in the iOS app, as VLCKit does the correct thing, as far as I can tell.

carolanitz:
The freaking VLCEventManager! I tell you I've been eyeing that thing for a while. It swallows events every now and then and baaaaaaadly needs a rewrite. I just haven't gotten around to it. *angrily shaking my fist at code*

mkchoi212:
@carolanitz ohhh can I give `VLCEventManager` a go? 😎

carolanitz:
how about updating this PR first to use _mediaPlayer.position ;) ? Even if it sometimes doesn't work we should use this api here and fix what is broken underneath seperately

mkchoi212:
@carolanitz Oh yeah for sure

SharedSources/VLCRemoteControlService.m
```diff
@@ -32,7 +32,8 @@ @implementation VLCRemoteControlService
     if (@available(iOS 9.1, *)) {
         [commands addObject:cc.changePlaybackPositionCommand];
     }
-    return commands;
+    NSArray *finalCommands = [commands copy];
+    return finalCommands;
```
Mikanbu: I think you should directly `return [commands copy];` without creating a variable for it.

carolanitz:
just tried it !! This is awesome 👍 so cool this is working right now! This was something @fkuehne 
 and I wanted a long long time ago 🎉https://openradar.appspot.com/radar?id=5799386160300032

SharedSources/VLCRemoteControlService.m
```diff
@@ -125,6 +129,11 @@ - (MPRemoteCommandHandlerStatus )remoteCommandEvent:(MPRemoteCommandEvent *)even
         [_remoteControlServiceDelegate remoteControlService:self setPlaybackRate:rateEvent.playbackRate];
         return MPRemoteCommandHandlerStatusSuccess;
     }
+    if (event.command == cc.changePlaybackPositionCommand) {
```
fkuehne: This will raise a selector not found exception on iOS 9.0.

carolanitz:
this was merged

PR #10, opened 2018-03-05
VLCPlaybackController: fix openURL crash

carolanitz:
openURL with options and completion is iOS 10+ hence we're going back to the old call


carolanitz:
@dcordero since I can't pick you from the reviewer list (yet) I will mention you in further PRs ;)

carolanitz:
#merged

PR #11, opened 2018-03-05
fix crash on bad url

carolanitz:
We're also showing an alert when the user enters a wrong url now

### DIFF
--- a/Sources/VLCOpenNetworkStreamViewController.m
+++ b/Sources/VLCOpenNetworkStreamViewController.m
@@ -178,22 +178,33 @@
 
 - (IBAction)openButtonAction:(id)sender
 {
-    if ([self.urlField.text length] > 0) {
-        if (!self.privateToggleSwitch.on) {
-            NSString *urlString = self.urlField.text;
-            if ([_recentURLs indexOfObject:urlString] != NSNotFound)
-                [_recentURLs removeObject:urlString];
+    if ([self.urlField.text length] <= 0 || [NSURL URLWithString:self.urlField.text] == nil) {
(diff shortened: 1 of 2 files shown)


carolanitz:
@dcordero :)

carolanitz:
btw this can be reproduced by entering an emoji 💃into the url textfield

Sources/VLCOpenNetworkStreamViewController.m
```diff
@@ -178,70 +178,66 @@ - (IBAction)goBack:(id)sender
...
+    UIAlertController *alertController = [UIAlertController alertControllerWithTitle:NSLocalizedString(@"BUTTON_RENAME", nil)
+                                                                             message:nil
+                                                                      preferredStyle:UIAlertControllerStyleAlert];
+
+    UIAlertAction *cancelAction = [UIAlertAction actionWithTitle:NSLocalizedString(@"BUTTON_CANCEL", nil)
+                                                           style:UIAlertActionStyleCancel
+                                                         handler:nil];
+    UIAlertAction *okAction = [UIAlertAction actionWithTitle:@"OK" style:UIAlertActionStyleDefault handler:^(UIAlertAction * _Nonnull action) {
+        NSString *streamTitle = alertController.textFields.firstObject.text;
+        [self renameStreamWithTitle:streamTitle atIndex:cellIndexPath.row];
```
Mikanbu: I don't think we need `streamTitle`, we could directly use `alertController.textFields.firstObject.text` but it could be debatable readability-wise 😄

carolanitz: yes I think it's actually nicer with the streamTitle readability wise and it provides the context that it is the new streamtitle

PR #18, opened 2018-03-16
Pods: added Pods since it is good practice

carolanitz:
Cocoapods recommends this and CircleCi does as well. It will also make sure that we can check out old versions and reproduce issues without worrying that the used pods are not available anymore
https://guides.cocoapods.org/using/using-cocoapods#should-i-check-the-pods-directory-into-source-control


Mikanbu:
I agree that it would bring great benefits, but for VLC-iOS in particular wouldn't it make the git repository too big since some of the pods we use are quite big?

thresheek:
FYI, this is what I received from github today:

Git LFS has been disabled on the organization videolan because you’ve exceeded your data plan by at least 150%. Please purchase additional data packs to cover your bandwidth and storage usage:

Current usage as of 16 Mar 2018 02:21AM UTC:

  Bandwidth: 0.0 GB / 1 GB (0%)
  Storage: 1.85 GB / 1 GB (185%)

carolanitz:
@Mikanbu yes it makes it very big. I thought about it too but figured since the user can't do anything if he doesn't install the pods immediately after checkout, why not check them in already. 
@thresheek how much is additional storage?

carolanitz:
not gonna do this. VLCKit is indeed too big and will blow up the repository. We could add VLCKit and TVVLCKit instead as submodules if we want to achieve reproducibility

thresheek:
Pricing is at https://github.com/pricing, and we're not going to pay for github, sorry.

carolanitz:
I could've/ should've googled that :D but yeah I never expected us to pay but was curious and after discussing that further with someone it's anyway not the best idea :)

PR #20, opened 2018-03-21
Small fixes regarding Cocoapods 🍫

dcordero:
- [x] Organize Podfile to avoid repeting the same dependencies among different targets
- [x] Update README instructions to build the project. It is not requiered to update dependencies, but only to install them


fkuehne:
True, you are just changing that in the Podfile.lock, not in the actual Podfile. Regarding the watchOS target, IIRC, I need to tweak something in the MLKit project first before we can use the prod variant there.

PR #21, opened 2018-03-22
[Doc] Add new documentation regarding beginner issues

mkchoi212:
When I first decided to contribute to VLC-iOS, I had some trouble building and running the app. After receiving help from the community (@carolanitz), I got it to build but having a guide such as this could have save me a lot of time. 

And because I couldn't find anything in the repo similar to this, I thought I should create a guide of some sort. 😃

Related Issue : https://code.videolan.org/videolan/vlc-ios/issues/129.


fkuehne:
This should be part of https://wiki.videolan.org/iOSCompile and not of a readme stored within the source code.

carolanitz:
This is great to have and just needs to go somewhere else 👍 let us know if you need help with editing the wiki

mkchoi212:
Currently trying to create an account in the wiki from link below
> To modify the contents of this wiki, you need to [create an account](https://wiki.videolan.org/Special:CreateAccount/), which is very fast

But I'm getting the following message

> You do not have permission to create this user account, for the following reason:
> The action you have requested is limited to users in the group: Administrators.

PR #23, opened 2018-03-22
VLCOneDrive: fix tvOS build

carolanitz:
UIAlertView is not available on tvOS so we need to use the UIAlertController instead


Mikanbu:
This was merged!

PR #24, opened 2018-03-23
tvOS: Fix play after pause with Siri Remote

Mikanbu:
This fixes the the play functionality that was blocked after a pause.


carolanitz:
merged

PR #25, opened 2018-03-24
[Playback] Update position in Control Center

mkchoi212:
Previously, when position was set for a media player, only its metadata was updated.
Now, its position is updated correspondingly for Control Center.

In detail,
- When opening previously played file, its position is now reflected in Control Center
- Changing playback position using scrubber in the app's player is reflected in Control Center

Thought something was wrong with `VLCEventManager.m` initially, but it seems like all event messages are being processed correctly. @carolanitz


Sources/VLCPlaybackController.m
```diff
@@ -516,6 +516,8 @@ - (float)playbackPosition
 - (void)setPlaybackPosition:(float)position
 {
     _mediaPlayer.position = position;
+    float duration = self.mediaDuration / 1000.;
+    [self remoteControlService:_remoteControlService setCurrentPlaybackTime:duration * position];
```
carolanitz: I think calling this just triggers an update through the jumpforward call but the bug is that setting the position doesn't trigger an update in VLCMetadata. calling self.setsmetadataupdate should fix this. Could you try that instead ?

mkchoi212: Ok so the weirdest thing is happening. It's as if the requests for updating the infocenter with meta data is being queued up.

E.g. If I first open a file, which I left off at 1:31, infocenter starts at 0:00. If I move the slider to 2:00, infocenter gets updated to 1:31.

carolanitz: weird. This is probably our bug. it reads out the old position when you update i think.

mkchoi212: I'm kind of running out of ideas how to debug this :( Tried looking at the stack trace but can't really do much from a recorded stack frame. And I'm not even really sure where the bug is

PR #27, opened 2018-03-31
Add Ruby and Swift linters and PR template

dcordero:
- [x] Add Swift and Ruby linters
- [x] Define PR template for GitHub
- [x] Define lint lane
- [x] Add lint step to circleci
- [x] Fix linter issues in Ruby and Swift code


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
+<!-- Thanks for contributing to _vlc-ios_! Before you submit your pull request, please make sure to check the following boxes by putting an x in the [ ] (don't: [x ], [ x], do: [x]) -->
+
+### Checklist
+- [ ] I've run `bundle exec fastlane test` from the root directory to see all new and existing tests pass
```
carolanitz: How do you feel about making this part of circle ci and adjust the text?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,10 @@
...
+- [ ] I've followed the vlc-ios code style and run `bundle exec fastlane lint` to ensure the code style is valid
```
carolanitz: can we adjust the code style.rtf (which probably should turn into an .md file) and link to it here ? :)

.circleci/config.yml
```diff
@@ -20,6 +20,7 @@ jobs:
             - .bundle
       - restore_cache:
           key: v1-pods-{{ checksum "Podfile.lock" }}
+      - run: brew install swiftlint
```
carolanitz: I think this might also need to go into the README for requirements :)

fastlane/Fastfile
```diff
@@ -69,28 +76,27 @@ end
...
 private_lane :update_changelog do |options|
+  platform = options[:platform]
   # Splits the News by -------- get out the top notes
   changelog = File.read('../Docs/NEWS').split('-----------')[1].split('-----------').first
-  temp_changelog = changelog.split("${options[:platform]}")
-  if temp_changelog.count <= 1
-    temp_changelog = changelog.split("tvOS")
-  end
+  temp_changelog = changelog.split(platform)
+  temp_changelog = changelog.split('tvOS') if temp_changelog.count <= 1
```
carolanitz: nice! I didn't know that's how you can write this

.circleci/config.yml
dcordero: Good point, I have opted in by using swiftlint via Cocoapods so it does not introduce additional steps.

fastlane/Fastfile
dcordero: Me neither tbh, rubocop made this change 😆

.github/PULL_REQUEST_TEMPLATE.md
dcordero: I could not agree more... done ✅

fastlane/Fastfile
```diff
@@ -41,6 +41,7 @@ end
 
 lane :ci do
   lint
+  test
```
carolanitz: I would move the running of tests step after the building step :)

Docs/Objc_coding_style.md
```diff
@@ -0,0 +1,158 @@
+# Objective-c code style
+
+Variable names should use capitalization on all word expect the first, and never '_' ie
+
+**Right:**
+```obj-c
+BOOL isFirstTimeReading;
+int age;
+Array *myArray;
```
carolanitz: NSArray ? :D

dcordero: Good catch

fastlane/Fastfile
dcordero: Ok, I do not have a strong opinion about that. Changed ✅ 

In fact as soon as we have tests for both targets, the `test` step will make redundant the building steps, because it will build the App to run the apps.

But that is a next step...

carolanitz:
squashed and merged with fad3d6879c874d906d607d5974ea7b0a796cb54c

PR #28, opened 2018-04-01
Automated framed screenshots

mkchoi212:
Major feat in this PR are
- Support for `fastlane screenshots` *with localization*
- New UI tests for the new shiny UI  *with localization*

I wasn't sure which screens to capture and what kind of text to put in the frames but here's what I came up with. Here's some demo screenshots in German; thanks to Google Translate 😃

> Please note that code for screenshooting is part of UI testing


mkchoi212:
circleci seems to be getting the following error, which I'm not getting on my local machine
```
❌  /Users/distiller/project/VLC for iOSUITests/VLCiOSTestVideoCodecs.swift:13:8: cannot load underlying module for 'XCTest'
[00:21:20]: ▸ import XCTest
```

dcordero:
Great work, the screenshots looks awesome !!!

I have added a few comments here and there

Podfile
```diff
@@ -23,6 +23,7 @@ def iOS_pods
   pod 'MediaLibraryKit-prod'
   pod 'MobileVLCKit', '3.0.2'
   pod 'GTMAppAuth'
+  pod 'SimulatorStatusMagic', :configurations => ['Debug']
```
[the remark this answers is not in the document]
mkchoi212: Moving `pod 'SimulatorStatusMagic'` to `target 'VLC for iOSUITests'` won't allow `VLCAppDelegate.m`to import the module, right?

https://github.com/mkchoi212/vlc-ios/blob/6383eee5f46849f493a4ee29e6b7e9224f2c0622/Sources/VLCAppDelegate.m#L38

dcordero:
The error in circleci means that XCTest is been imported from an Application target. Are the those new swift files included in the test target or the app target? That could be the reason of the error.

VLC for iOSUITests/VLCiOSTestMenu.swift
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import Foundation
+import XCTest
+
+class VLCiOSTestMenu: XCTestCase {
+    let app = XCUIApplication()
+    let moreTab = XCUIApplication().tabBars.buttons.element(boundBy: 4)
```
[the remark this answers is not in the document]
mkchoi212: There is no localizable string for this :(
I think the Video tab is using Apple's built-in localized strings.

VLC for iOSUITests/VLCiOSTestMenu.swift
```diff
@@ -0,0 +1,93 @@
...
+    func testNavigationToNetworkTab() {
+        let localNetwork = localizedString(key: "LOCAL_NETWORK")
+        app.tabBars.buttons[localNetwork].tap()
+        XCTAssertNotNil(app.navigationBars[localNetwork])
+        
+        snapshot("network_tab")
+    }
+    
+    func testNavigationToVideoTab() {
+        app.tabBars.buttons["Video"].tap()
```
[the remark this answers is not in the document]
mkchoi212: There is no localizable string for this either :(

.gitignore
```diff
@@ -10,7 +10,6 @@ xcuserdata
 # Fastlane
 fastlane/report.xml
 fastlane/Preview.html
-fastlane/screenshots
```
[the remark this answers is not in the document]
mkchoi212: I will make modifications based on [this](https://github.com/fastlane/fastlane/blob/bfc7a6e277bbede8b15c1d89a31ebdd7c5e90b63/docs/Advanced.md#gitignore) guide! 👍

fastlane/Snapfile
```diff
@@ -0,0 +1,13 @@
+ devices([
+   "iPhone X",
+   "iPhone 8"
+ ])
+
+ languages([
+   "en",
+   "de",
+   "fr",
```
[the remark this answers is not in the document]
mkchoi212: To create the framed screenshots, we need localized `keyword.string` and `title.string` for each language. Maybe this should be a progressive effort once we officially decide on what text to put on the framed screenshots??

.gitignore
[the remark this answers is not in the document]
carolanitz: awesome that's even better! Great find :)

VLC for iOSUITests/VLCiOSTestMenu.swift
```diff
@@ -0,0 +1,81 @@
...
+    let app = XCUIApplication()
+    var helper: TestHelper!
+
+    override func setUp() {
+        super.setUp()
+
+        XCUIDevice.shared.orientation = .portrait
+        setupSnapshot(app)
+        helper = TestHelper(lang: deviceLanguage, target: VLCiOSTestMenu.self)
+        app.launch()
```
carolanitz: This here would actually be a better place for the status magic (but remember to disable it again in tearDown()).

VLC for iOSUITests/TestHelper.swift
```diff
@@ -0,0 +1,58 @@
...
+ * Authors: Mike JS. Choi <mkchoi212 # icloud.com>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import Foundation
+import XCTest
+
+struct TestHelper {
+    let localizationBundle: Bundle
```
carolanitz: something that I didn't think about in my earlier review but that a friend pointed out is to use accessibility element identifiers. They don’t need to be localized! Since we need to make the app accessibility friendly anyway this will kill two birds with one stone :)

VLC for iOSUITests/VLCiOSTestMenu.swift
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    func testNavigationToAudioTab() {
+        let audio = helper.localized(key: "AUDIO")
+        helper.tap(tabDescription: audio, app: app)
+        XCTAssertNotNil(app.navigationBars[audio])
+
+        snapshot("audio_tab")
```
carolanitz: btw could we separate the Unit Tests from the snapshot taking?

fastlane/Snapfile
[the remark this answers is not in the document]
carolanitz: absolutely let's just ask Louis!

VLC for iOSUITests/TestHelper.swift
mkchoi212: So it looks like none of the elements in the app has its `accessibilityIdentifier` set. How should we go about this? Update every single element's `accessibilityIdentifier`?

carolanitz: yes exactly. It needs to be added

.gitignore
```diff
@@ -10,9 +10,11 @@ xcuserdata
...
 fastlane/test_output
 
+fastlane/screenshots/*/*.png
+fastlane/screenshots/screenshots.html
```
carolanitz: will you hate me if I ask for an alphabetically ordered .gitignore 🙊?

mkchoi212: Not at all 😄

carolanitz:
looks already better. Could you use the accessibilityidentifier on the tabbaritems and other views as well? We need to avoid all use of the localization bundle otherwise renaming a tab would lead to failing tests for no reason

carolanitz:
There is also something misconfigured in the xcodeproj for me. Might be just me though screen shot 2018-04-09 at 2 16 01 pm

fastlane/Snapfile
```diff
@@ -0,0 +1,31 @@
...
+   "iPhone SE", 
+   "iPhone X", 
+   "iPad Air", 
+   "iPad Air 2", 
+   "iPad (5th generation)", 
+   "iPad Pro (9.7-inch)", 
+   "iPad Pro (12.9-inch)", 
+   "iPad Pro (12.9-inch) (2nd generation)", 
+   "iPad Pro (10.5-inch)"
+ ])
```
carolanitz: 👍

carolanitz:
the tests fail on iPad btw but the more tab will disappear anyway soon. And one more question. What are the screenshots doing for a language that is not defined? can we make sure that it uses the screenshots from the English version?

mkchoi212:
@carolanitz Right now, `frameit`does not work with any language that does not have their respective `.string` files. But let me see if I can make the default language English for them!

mkchoi212:
Please note that because the "More" tab is going to be removed soon, the test does not take the "More" tab into consideration. Because of this, **the tests fail for all iPhones but succeed for all iPads**

carolanitz:
the merge conflicts just need to be resolved and I think this is good to go

mkchoi212:
Fixed all the merge conflicts! 
Sorry for such a huge PR but thanks for the comments @dcordero @carolanitz ❤️
I'll make sure to break up these bad boys into smaller chunks next time!

mkchoi212:
All commits have been squashed!

VLC.xcodeproj/project.pbxproj
```diff
@@ -3431,7 +3408,6 @@
 				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
 				CLANG_ANALYZER_NONNULL = YES;
 				CLANG_ANALYZER_NUMBER_OBJECT_CONVERSION = YES_AGGRESSIVE;
-				CLANG_CXX_LANGUAGE_STANDARD = "gnu++14";
```
mkchoi212: @carolanitz, this fixes the misconfigured in the xcodeproj problem that you discovered!

PR #29, opened 2018-04-09
Empty view for media library view

mkchoi212:
Nothing much to it!
Just displays the empty view when the media controller is empty 😄
Please note that I made some adjustments to the UI as I saw fit to the new UI.

### DIFF
--- a/SharedSources/MediaDataSource.swift
+++ b/SharedSources/MediaDataSource.swift
@@ -28,10 +28,7 @@ public class MediaDataSourceAndDelegate: NSObject, UICollectionViewDataSource, U
     }
 
     public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        let numItems = Int(services.mediaDataSource.numberOfFiles())
-        let hasMedia = numItems > 0
-        collectionView.backgroundView?.isHidden = hasMedia
-        return numItems
+        return Int(services.mediaDataSource.numberOfFiles())
     }
 
     public func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
--- a/SharedSources/MediaDataSource.swift
+++ b/SharedSources/MediaDataSource.swift
@@ -6,6 +6,7 @@
  * $Id$
  *
  * Authors: Carola Nitz <nitz.carola # gmail.com>
+ *          Mike JS. Choi <mkchoi212 # icloud.com>
  *
  * Refer to the COPYING file of the official project for license.
  *****************************************************************************/
@@ -27,7 +28,10 @@ public class MediaDataSourceAndDelegate: NSObject, UICollectionViewDataSource, U
     }
 
     public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        return Int(services.mediaDataSource.numberOfFiles())
+        let numItems = Int(services.mediaDataSource.numberOfFiles())
+        let hasMedia = numItems > 0
+        collectionView.backgroundView?.isHidden = hasMedia
+        return numItems
     }
 
     public func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
--- a/Sources/MediaViewController.swift
+++ b/Sources/MediaViewController.swift
@@ -6,6 +6,7 @@
  * $Id$
  *
  * Authors: Carola Nitz <nitz.carola # gmail.com>
+ *          Mike JS. Choi <mkchoi212 # icloud.com>
  *
  * Refer to the COPYING file of the official project for license.
  *****************************************************************************/
@@ -29,6 +30,17 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         let dragAndDropManager = VLCDragAndDropManager()
         dragAndDropManager.delegate = services.mediaDataSource
         return dragAndDropManager
+    }()
+    
+    lazy var emptyView: VLCEmptyLibraryView = {
+        let name = String(describing: VLCEmptyLibraryView.self)
+        let nib = Bundle.main.loadNibNamed(name, owner: self, options: nil)
+        guard let emptyView = nib?.first as? VLCEmptyLibraryView else {
+            assertionFailure("Can't find nib for \(name)")
+            return VLCEmptyLibraryView()
+        }
+        emptyView.frame = view.bounds
+        return emptyView
     }()
 
     public convenience init(services: Services) {
@@ -76,12 +88,13 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
 
     override public func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+        displayEmptyViewIfNeeded()
         services.mediaDataSource.updateContents(forSelection: nil)
         services.mediaDataSource.addAllFolders()
         services.mediaDataSource.addRemainingFiles()
         collectionView?.reloadData()
-
     }
+    
     func setupSearchController() {
         searchController = UISearchController(searchResultsController: nil)
         searchController?.searchResultsUpdater = self
@@ -104,6 +117,10 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
 
     func setupNavigationBar() {
         navigationItem.leftBarButtonItem = UIBarButtonItem(title: NSLocalizedString("SORT", comment: ""), style: .plain, target: self, action: #selector(sort))
+    }
+    
+    func displayEmptyViewIfNeeded() {
+        collectionView?.backgroundView = collectionView?.numberOfItems(inSection: 0) == 0 ? emptyView : nil
     }
 
     @objc func sort() {
--- a/Sources/VLCEmptyLibraryView.m
+++ b/Sources/VLCEmptyLibraryView.m
@@ -1,0 +1,35 @@
+/*****************************************************************************
+ * VLCEmptyLibraryView.m
+ * VLC for iOS
+ *****************************************************************************
+ * Copyright (c) 2018 VideoLAN. All rights reserved.
+ * $Id$
+ *
+ * Authors: Mike JS. Choi <mkchoi212 # icloud.com>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+#import <Foundation/Foundation.h>
+#import "VLCEmptyLibraryView.h"
+#import "VLCFirstStepsViewController.h"
+
+@implementation VLCEmptyLibraryView
+
+- (void)awakeFromNib
+{
+    _emptyLibraryLabel.text = NSLocalizedString(@"EMPTY_LIBRARY", nil);
+    _emptyLibraryLongDescriptionLabel.text = NSLocalizedString(@"EMPTY_LIBRARY_LONG", nil);
+    [_learnMoreButton setTitle:NSLocalizedString(@"BUTTON_LEARN_MORE", nil) forState:UIControlStateNormal];
+    [super awakeFromNib];
+}
+
+- (IBAction)learnMore:(id)sender
+{
+      UIViewController *firstStepsVC = [[VLCFirstStepsViewController alloc] initWithNibName:nil bundle:nil];
+      UINavigationController *navCon = [[UINavigationController alloc] initWithRootViewController:firstStepsVC];
+      navCon.modalPresentationStyle = UIModalPresentationFormSheet;
+      [self.window.rootViewController presentViewController:navCon animated:YES completion:nil];
+}
+
+@end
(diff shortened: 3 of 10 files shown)


Sources/VLCLibraryViewController.m
```diff
@@ -39,6 +40,20 @@
 
 @implementation EmptyLibraryView
 
+- (id)initWithFrame:(CGRect)frame isInFolder:(bool)inFolder
```
carolanitz: The inFolder bool can go, we won't have folders anymore

Sources/VLCLibraryViewController.m
```diff
@@ -39,6 +40,20 @@
...
+{
+    if (self) {
+        NSArray *subviewArray = [[NSBundle mainBundle] loadNibNamed:@"VLCEmptyLibraryView" owner:self options:nil];
+        self = [subviewArray objectAtIndex:0];
```
carolanitz: you should always use .firstObject because it will give you nil if there are not subviews.
Right now this would crash if something goes wrong with loading the Nib

SharedSources/MediaDataSource.swift
```diff
@@ -27,7 +28,13 @@ public class MediaDataSourceAndDelegate:NSObject, UICollectionViewDataSource, UI
     }
 
     public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        return Int(services.mediaDataSource.numberOfFiles())
+        let numItems = Int(services.mediaDataSource.numberOfFiles())
+        if numItems == 0 {
```
carolanitz: can we write 
Bool hasMedia = numItems > 0
collectionView.backgroundView?.isHidden = hasMedia

Sources/VLCLibraryViewController.m
carolanitz: This doesn't look correct to me. Shouldn't it be if (!self) and then you initialize? I need to take another look but it feels to me like this code doesn't belong here as well

Hruks: My 5 cents:
Usually override of init require call 
self = [super init];
In this case self will be setup properly and we need to check for "if (self)".
In case if we need load view from xib file the common solution is to make static method to do this. For example:
```
+ (instanceType)emptyLibraryViewInFolder:(bool)inFolder
{
    NSArray *subviewArray = [[NSBundle mainBundle] loadNibNamed:@"VLCEmptyLibraryView" owner:self options:nil];
    EmptyLibraryView *result = subviewArray.firstObject;
    [result setupEmptyValuesInFolder:inFolder];
    return result;
}

- (void)setupEmptyValuesInFolder:(bool)inFolder
{
    _emptyLibraryLabel.text = inFolder ? NSLocalizedString(@"FOLDER_EMPTY", nil) : NSLocalizedString(@"EMPTY_LIBRARY", nil);
    _emptyLibraryLongDescriptionLabel.text = inFolder ? NSLocalizedString(@"FOLDER_EMPTY_LONG", nil) : NSLocalizedString(@"EMPTY_LIBRARY_LONG", nil);
    [_learnMoreButton setTitle:NSLocalizedString(@"BUTTON_LEARN_MORE", nil) forState:UIControlStateNormal];
    _learnMoreButton.hidden = inFolder;
}

```

carolanitz: thumbs up for the mentioning of super

Sources/VLCLibraryViewController.m
mkchoi212: Can I go ahead and delete all localized strings for `FOLDER_EMPTY` and `FOLDER_EMPTY_LONG`?

carolanitz: absolutely

Sources/MediaViewController.swift
```diff
@@ -53,6 +54,7 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         setupCollectionView()
         setupSearchController()
         setupNavigationBar()
+        setUpEmptyBackgroundView()
```
dcordero: `setUp` vs `setup`

I do not have any preference among them, but just because of karma I think they should be all the same.

mkchoi212: Oh I didn't even notice that. Thanks @dcordero

SharedSources/MediaDataSource.swift
```diff
@@ -27,7 +28,10 @@ public class MediaDataSourceAndDelegate:NSObject, UICollectionViewDataSource, UI
     }
 
     public func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        return Int(services.mediaDataSource.numberOfFiles())
+        let numItems = Int(services.mediaDataSource.numberOfFiles())
+        let hasMedia = numItems > 0
+        collectionView.backgroundView?.isHidden = hasMedia
```
carolanitz: This code doesn't belong here. A view should never be modified from a model class

VLC-iOS-Bridging-Header.h
```diff
@@ -9,6 +9,7 @@
 #import "VLCConstants.h"
 #import "VLCDownloadViewController.h"
 #import "VLCLibrarySearchDisplayDataSource.h"
+#import "VLCLibraryViewController.h"
```
carolanitz: Why do you need the LibraryViewController?

Apple-TV/VLCServerBrowsingTVViewController.m
```diff
@@ -48,7 +48,7 @@ - (void)viewDidLoad
 {
     [super viewDidLoad];
 
-    self.nothingFoundLabel.text = NSLocalizedString(@"FOLDER_EMPTY", nil);
```
carolanitz: Isn't that code related to network browsing on the Apple TV?  If so then we still have folders here and this code needs to stay. I also wouldn't change any AppleTV related code in this PR

Sources/VLCLibraryViewController.m
```diff
@@ -406,8 +406,8 @@ - (void)_displayEmptyLibraryViewIfNeeded
 
     if ([_mediaDataSource numberOfFiles] == 0) {
         _inFolder = (_libraryMode == VLCLibraryModeFolder || _libraryMode == VLCLibraryModeCreateFolder);
```
carolanitz: the VLCLibrarycontroller doesn't need to be touched here. It's gonna be deleted anyway

VLC-iOS-Bridging-Header.h
mkchoi212: `VLCLibraryViewController.h` is where `EmptyLibraryView` is declared and used within `VLCLibraryViewController.h`. Maybe writing a new version of it in Swift is beneficial in the long run?

Mikanbu: Yes I think it should be taken out since `VLCLibraryViewController` will be removed "soon".

carolanitz: It shouldn't have been declared in the LibraryViewController in the first place. It needs to live in it's own file. You don't have to write a new version in Swift, just move it into it's own class via copy paste :)

Sources/VLCLibraryViewController.h
```diff
@@ -37,6 +37,7 @@
 @property (nonatomic, strong) IBOutlet UIActivityIndicatorView *activityIndicator;
 @property (nonatomic, strong) IBOutlet UIButton *learnMoreButton;
 
+- (id)initWithFrame:(CGRect)frame;
```
carolanitz: doesn't need to be defined since EmptyView is a UIView subclass

Sources/VLCLibraryViewController.m
```diff
@@ -39,6 +40,20 @@
 
 @implementation EmptyLibraryView
 
+- (id)initWithFrame:(CGRect)frame
+{
+    self = [super initWithFrame:frame];
+    if (self) {
+        NSArray *subviewArray = [[NSBundle mainBundle] loadNibNamed:@"VLCEmptyLibraryView" owner:self options:nil];
```
carolanitz: initiWithFrame is actually the wrong place to load from the nib. You should take a look how it's done everywhere else. You're also calling super and assigning it to self, then checking for nil and then assigning self = subviewArray.firstObject; without having the nil check again. I missed that the first time this needs to be done differently

SharedSources/MediaDataSource.swift
carolanitz: we should also consider not just hiding the emptyview but consider removing it. It might lead to side effects having this view still in the view hierarchy

Sources/MediaViewController.swift
```diff
@@ -105,6 +109,19 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
     func setupNavigationBar() {
         navigationItem.leftBarButtonItem = UIBarButtonItem(title: NSLocalizedString("SORT", comment: ""), style: .plain, target: self, action: #selector(sort))
     }
+    
+    func setupEmptyBackgroundView() {
+        emptyView = VLCEmptyLibraryView.init()
```
carolanitz: you shouldn't have the init() here and should instantiate the View here from the nib

Sources/MediaViewController.swift
```diff
@@ -19,6 +20,7 @@ import Foundation
 
 public class VLCMediaViewController: UICollectionViewController, UISearchResultsUpdating, UISearchControllerDelegate {
     private var services: Services
+    private var emptyView: VLCEmptyLibraryView?
```
carolanitz: I think you can use a lazy var here and do the instantiation here

Sources/MediaViewController.swift
```diff
@@ -105,6 +109,19 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
...
         navigationItem.leftBarButtonItem = UIBarButtonItem(title: NSLocalizedString("SORT", comment: ""), style: .plain, target: self, action: #selector(sort))
     }
+    
+    func setupEmptyBackgroundView() {
+        emptyView = VLCEmptyLibraryView.init()
+        emptyView?.frame = view.bounds
+    }
+    
+    func displayEmptyViewIfNeeded() {
+        guard let view = emptyView, collectionView?.numberOfItems(inSection: 0) == 0 else {
```
carolanitz: since the emptyview will be lazy it's not optional anymore and you can simply write 
collectionView?.backgroundView = collectionView?.numberOfItems(inSection: 0) == 0 ? emptyView : nil

Sources/VLCEmptyLibraryView.m
```diff
@@ -0,0 +1,41 @@
...
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+#import <Foundation/Foundation.h>
+#import "VLCEmptyLibraryView.h"
+#import "VLCFirstStepsViewController.h"
+
+@implementation VLCEmptyLibraryView
+
++ (instancetype)init
```
carolanitz: This should probably be initWithCoder and you shouldn't have an instance of VLCEmptyView in it's own init, also no asserts but give the view in the xib an identifier and you load the nib with identifier

carolanitz: that should remove the whole looking through subviews and you just got one instance but this code belongs in the mediaviewcontroller. Also New Classes should be in Swift 😀

Sources/VLCEmptyLibraryView.m
```diff
@@ -0,0 +1,41 @@
...
+{
+    NSArray *subviews = [[NSBundle mainBundle] loadNibNamed:@"VLCEmptyLibraryView" owner:self options:nil];
+    NSAssert([subviews count] == 1, @"VLCEmptyLibraryView nib should contain only one subview");
+    NSAssert([[subviews firstObject] isKindOfClass:[VLCEmptyLibraryView class]], @"Loaded VLCEmptyLibraryView nib is invalid");
+    
+    VLCEmptyLibraryView *view = subviews.firstObject;
+    view.emptyLibraryLabel.text = NSLocalizedString(@"EMPTY_LIBRARY", nil);
```
carolanitz: you should move this code in override awakeFromNib()

Sources/MediaViewController.swift
```diff
@@ -30,6 +31,14 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         dragAndDropManager.delegate = services.mediaDataSource
         return dragAndDropManager
     }()
+    
+    lazy var emptyView: VLCEmptyLibraryView = {
+        let name = String(describing: VLCEmptyLibraryView.self)
+        let nib = Bundle.main.loadNibNamed(name, owner: self, options: nil)
+        guard let emptyView = nib?.first as? VLCEmptyLibraryView else { fatalError("Can't find nib for \(name)") }
```
carolanitz: I don't think we need to crash the app if we can't display an emptyview. An assertion and returning a UIView would've been just as fine

carolanitz:
merged with 6c1b90990e13467626c61bdcfb630dac395a6837 & fbcf87b45126b898e373f53b0ba33b1a1d3cfb82

PR #31, opened 2018-04-10
Remove Watchcode and related classes

carolanitz:
Since the watch is essentially a bigger remote at this point and sadly also not well maintained, we decided to remove it and shift focus to the iOS and tvOS app. Should the watch gain the ability to play video we will consider adding it back


Podfile
```diff
@@ -31,12 +31,6 @@ target 'VLC-iOS' do
   iOS_pods
```
dcordero: Due to the fact that `iOS_pods` are now only used by this single target, I think having those pods defined in a external function is not longer needed and they could be all moved to the `VLC-iOS` target itself

dcordero:
A couple of things regarding Fastlane configuration, could be also removed:

```diff
diff --git a/fastlane/Fastfile b/fastlane/Fastfile
index 36feb2e5..f3879bfa 100644
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -73,11 +73,7 @@ private_lane :set_version_bump_build_and_commit do |options|
     set_version_number_in_plist(VLC::infoPlistPath[:tvOS], options[:version])
   elsif options[:platform] == 'iOS'
     increment_build_number_in_plist(VLC::infoPlistPath[:iOS])
-    increment_build_number_in_plist(VLC::infoPlistPath[:watchKitExtension])
-    increment_build_number_in_plist(VLC::infoPlistPath[:watchOS])
     set_version_number_in_plist(VLC::infoPlistPath[:iOS], options[:version])
-    set_version_number_in_plist(VLC::infoPlistPath[:watchKitExtension], options[:version])
-    set_version_number_in_plist(VLC::infoPlistPath[:watchOS], options[:version])
   end
```

and also:

```diff
diff --git a/fastlane/helpers/VLC.rb b/fastlane/helpers/VLC.rb
index 25b80ef7..5a7c0321 100755
--- a/fastlane/helpers/VLC.rb
+++ b/fastlane/helpers/VLC.rb
@@ -5,8 +5,6 @@ class VLC
     {
       iOS: '../Sources/VLC for iOS-Info.plist',
       tvOS: '../Apple-TV/Info.plist',
-      watchKitExtension: '../VLC WatchKit Native Extension/Info.plist',
-      watchOS: '../VLC WatchKit Native/Info.plist'
     }
   end
 end
```

carolanitz:
merged with 368887bd83d5912ab8f17b089134dccfa42353de

PR #32, opened 2018-04-11
VLCAlertView refactor

carolanitz:
### Checklist

### Description

I'm submitting this on behalf of Pierre who took the first stab at replacing the old AlertView with the new Api :) This patch includes his very first lines of Swift code so let's rework this together!


Sources/VLCAlertControllerExtension.swift
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Authors: Pierre Sagaspe <pierre.sagaspe # me.com>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+typealias UIAlertCompletionBlock = (UIAlertController, Int) -> Void
+typealias UIAlertTextFieldCompletionBlock = (UIAlertController, Int, String) -> Void
+
+struct buttonKeys {
```
dcordero: This struct seems to be only used here internally, what do you think about defining it as `private`?

Anyway, since it defines a date type, I think it should be uppercased.

Sources/VLCAlertControllerExtension.swift
```diff
@@ -0,0 +1,109 @@
...
+    static var blocksDestructiveButtonIndex: Int = 2
+}
+
+@objc extension UIAlertController {
+
+    @objc static func showAlertInViewController( _ viewController: UIViewController,
+                                                            title: String,
+                                                          message: String? = nil,
+                                                cancelButtonTitle: String,
+                                                otherButtonTitles: String? = nil,
```
dcordero: If I it understand correctly, so far there support for an individual other button, should this parameter name be renamed to singular?

Sources/VLCLibraryViewController.m
```diff
@@ -766,12 +767,16 @@ - (void)collectionView:(UICollectionView *)collectionView requestToMoveItemAtInd
...
-        };
-        [alert show];
+        [UIAlertController showAlertInViewController:self
+                                               title:NSLocalizedString(@"FOLDER_INVALID_TYPE_TITLE", nil)
+                                             message:NSLocalizedString(@"FOLDER_INVALID_TYPE_MESSAGE", nil)
+                                   cancelButtonTitle:NSLocalizedString(@"BUTTON_OK", nil)
+                                   otherButtonTitles:nil
+                              destructiveButtonTitle:nil
+                                            tapBlock:^(UIAlertController *alertController, NSInteger buttonIndex) {
+                                                [self updateViewContents];
```
dcordero: This is a tricky one, but I think there a memory leak here due to a retain cycle 🤓

The tapBlock closure gets a strong reference to self (VLCLibraryViewController), so when UIAlertController is dismissed it will not be deallocated because of that reference.

We might need to get a weak reference to self instead.

```diff
+__weak typeof(self) weakSelf = self;
[UIAlertController showAlertInViewController:self
                                                title:NSLocalizedString(@"FOLDER_INVALID_TYPE_TITLE", nil)
                                              message:NSLocalizedString(@"FOLDER_INVALID_TYPE_MESSAGE", nil)
                                    cancelButtonTitle:NSLocalizedString(@"BUTTON_OK", nil)
                                    otherButtonTitles:nil
                               destructiveButtonTitle:nil
                                             tapBlock:^(UIAlertController *alertController, NSInteger buttonIndex) {
-                                                [self updateViewContents];
+                                                __strong typeof(weakSelf) strongSelf = weakSelf;
+                                                [strongSelf updateViewContents];
                                                [alertController dismissViewControllerAnimated:YES completion:nil];
                                             }];
```

dcordero:
Awesome work !!! 

I am really happy that we can get rid of a lot of warnings in the project because of using UIAlertView \o/.

carolanitz:
closing since this will be addressed by https://github.com/videolan/vlc-ios/pull/58

PR #38, opened 2018-04-19
VLCOpenNetworkStream: move controller from moretab to LocalNetwork tab

carolanitz:
### Checklist

### Description

This adjusts the tests as well for streaming. I have to bug louis though about this screen and the detailtext.
simulator screen shot - iphone 8 plus - 2018-04-19 at 17 20 03

### DIFF
--- a/Sources/LocalNetworkConnectivity/RemoteNetworkDataSource.swift
+++ b/Sources/LocalNetworkConnectivity/RemoteNetworkDataSource.swift
@@ -12,8 +12,14 @@
 import Foundation
 
 enum RemoteNetworkCellType: Int {
-    case cloud = 0, wifi
-    static let count: Int = 2
+    case cloud
+    case streaming
+    case wifi
+    static let count: Int = {
+        var max: Int = 0
+        while let _ = RemoteNetworkCellType(rawValue: max) { max += 1 }
+        return max
+    }()
 }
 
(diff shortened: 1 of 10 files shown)


Sources/LocalNetworkConnectivity/RemoteNetworkDataSource.swift
```diff
@@ -12,8 +12,8 @@
 import Foundation
 
 enum RemoteNetworkCellType: Int {
-    case cloud = 0, wifi
-    static let count: Int = 2
+    case cloud = 0, streaming, wifi
+    static let count: Int = 3
```
mkchoi212: Is the `cloud = 0`, required? 
I think enum automatically assigns the initial case's value to `0` :D

carolanitz: It's not automatically 0 from the documentation "Unlike C and Objective-C, Swift enumeration cases are not assigned a default integer value when they are created"

mkchoi212: Right! I think that only applies to when the enum's case has not been declared. If the type is declared,
> When you’re working with enumerations that store integer or string raw values, you don’t have to explicitly assign a raw value for each case. When you don’t, Swift will automatically assign the values for you.

PR #44, opened 2018-04-23
Chromecast integration

Mikanbu:
### Checklist

### Description

🚧 WIP Chromecast integration 🚧

~Should~ Include:
- Chromecast selection(MediaVC & MovieVC)
- Chromecast playback
- Chromecast change during playback
- Chromecast background playback
- Network disconnection handling
- Single Chromecast automatic connection on selection
- Audio priority with different apps(can cast and listen to Spotify~)

Still needs: 
- New mockup adaptation
- Plug with `PresentationTheme`
- Further testing with mulitple chromecast UI-wise
- ~Please not a~ Test suite


Sources/MediaViewController.swift
```diff
@@ -23,6 +23,8 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
     private var mediaDatasourceAndDelegate: MediaDataSourceAndDelegate?
     private var searchController: UISearchController?
     private let searchDataSource = VLCLibrarySearchDisplayDataSource()
+    private var rendererButtton: UIButton
```
fkuehne: tttypo.

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+        stackView.addArrangedSubview(name)
+        addSubview(stackView)
+
+        var guide: LayoutAnchorContainer = self
+
+        if #available(iOS 11.0, *) {
+            guide = safeAreaLayoutGuide
+        }
+        NSLayoutConstraint.activate([
+            icon.heightAnchor.constraint(equalToConstant: 25),
```
fkuehne: Maybe add a constant at the top of the file to easy future adaption of this constant?

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+            icon.widthAnchor.constraint(equalTo: icon.heightAnchor),
+
+            name.trailingAnchor.constraint(equalTo: stackView.trailingAnchor),
+            name.centerYAnchor.constraint(equalTo: stackView.centerYAnchor),
+
+            stackView.leadingAnchor.constraint(equalTo: guide.leadingAnchor, constant: 20),
```
fkuehne: idem.

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+}
+
+// MARK: VLCActionSheetSectionHeader
+class VLCActionSheetSectionHeader: UIView {
+
+    static let identifier = "VLCActionSheetSectionHeader"
+
+    let title: UILabel = {
+        let title = UILabel()
+        title.font = UIFont.systemFont(ofSize: 16)
```
fkuehne: idem.

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+@objc protocol VLCActionSheetDelegate: class {
+    @objc optional func headerViewTitle() -> String?
+    @objc func itemAtIndexPath(_ indexPath: IndexPath) -> Any?
+    @objc func actionSheet(collectionView: UICollectionView, didSelectItem item: Any, At indexPath: IndexPath)
+}
+
+// MARK: VLCActionSheet
+class VLCActionSheet: UIViewController {
+
+    private let cellHeight: CGFloat = 50
```
fkuehne: idem.

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+        return collectionView
+    }()
+
+    lazy var cancelButton: UIButton = {
+        let cancelButton = UIButton()
+        cancelButton.setTitle(NSLocalizedString("CANCEL", comment: ""), for: .normal)
+        cancelButton.setTitleColor(.white, for: .normal)
+        cancelButton.titleLabel?.font = UIFont.systemFont(ofSize: 16)
+        cancelButton.addTarget(self, action: #selector(self.removeActionSheet), for: .touchDown)
+        cancelButton.backgroundColor = UIColor(red:1.00, green:0.59, blue:0.13, alpha:1.0)
```
fkuehne: please use a global VLC color for this or create a new one.

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+        headerView.title.text = delegate?.headerViewTitle?() ?? "Default header title"
+        headerView.title.textColor = .black
+        headerView.backgroundColor = .white
+        headerView.translatesAutoresizingMaskIntoConstraints = false
+        return headerView
+    }()
+
+    lazy var bottomBackgroundView: UIView = {
+        let bottomBackgroundView = UIView()
+        bottomBackgroundView.backgroundColor = UIColor(red:1.00, green:0.59, blue:0.13, alpha:1.0)
```
fkuehne: idem.

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+
+    fileprivate lazy var maxCollectionViewHeightConstraint: NSLayoutConstraint = {
+        let maxCollectionViewHeightConstraint = collectionView.heightAnchor.constraint(
+            lessThanOrEqualToConstant: (view.bounds.height / 2) - cellHeight)
+        return maxCollectionViewHeightConstraint
+    }()
+
+    fileprivate lazy var collectionViewHeightConstraint: NSLayoutConstraint = {
+        guard let dataSource = dataSource else {
+            preconditionFailure("VLCActionSheet: Data source not setted correctly!")
```
fkuehne: s/setted/set

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+        // This is to avoid a horrible visual glitch!
+        mainStackView.isHidden = false
+
+        UIView.transition(with: backgroundView, duration: 0.2, options: .transitionCrossDissolve, animations: { [weak self] in
+            self?.backgroundView.isHidden = false
+            }, completion: nil)
+
+        let realMainStackView = mainStackView.frame
+
+        mainStackView.frame.origin.y += mainStackView.frame.origin.y
```
fkuehne: why? what? please add an explaining comment.

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+
+        UIView.animate(withDuration: 0.6, delay: 0, usingSpringWithDamping: 0.8, initialSpringVelocity: 0, options: .curveEaseOut, animations: {
```
fkuehne: Does it make sense to share this animation in a global way with more animated objects? See the shared animation in the tvOS interface.

Sources/VLCMovieViewController.m
```diff
@@ -1649,4 +1673,36 @@ - (void)handleExternalScreenDidDisconnect:(NSNotification *)notification
...
 }
 
+#pragma mark - Renderers
+
+- (void)playingExternallyViewWithRendererName:(NSString *)name
+{
+    // TODO: Have a better/prettier transition between local playback frame and playingExternallyView
+    _playingExternallyView.hidden = NO;
+    _playingExternallyTitle.text = NSLocalizedString(@"PLAYING_EXTERNALLY_TITLE_CHROMECAST", nil);
+    _playingExternallyTitle.text = @"Chromecast connected";
```
fkuehne: Missing l10n.

Sources/MediaViewController.swift
```diff
@@ -120,6 +135,11 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         collectionView?.backgroundView = collectionView?.numberOfItems(inSection: 0) == 0 ? emptyView : nil
     }
 
+    // MARK: Renderer
+    private func setupRendererDiscovererManager() {
```
carolanitz: the function name does not reflect what the method does

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+ * Authors: Soomin Lee <bubu@mikan.io>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import Foundation
+import UIKit
+
+// MARK: VLCRendererCollectionViewLayout
+class VLCRendererCollectionViewLayout: UICollectionViewFlowLayout {
```
carolanitz: This belongs in it's own class

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+        minimumInteritemSpacing = 0
+    }
+
+    override func shouldInvalidateLayout(forBoundsChange newBounds: CGRect) -> Bool {
+        return true
+    }
+}
+
+// MARK: VLCActionSheetCell
+class VLCActionSheetCell: UICollectionViewCell {
```
carolanitz: own class

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+        })
+    }
+
+    override func viewWillLayoutSubviews() {
+        super.viewWillLayoutSubviews()
+        collectionView.collectionViewLayout.invalidateLayout()
+    }
+
+    @objc func addAction(closure action: @escaping (_ item: Any) -> Void) {
+        self.action = action
```
carolanitz: it sets the action and doesn’t „add“ the action.

Sources/VLCMovieViewController.m
```diff
@@ -113,6 +114,9 @@ @interface VLCMovieViewController () <UIGestureRecognizerDelegate, VLCMultiSelec
     CGFloat _fov;
     CGPoint _saveLocation;
     CGSize _screenSizePixel;
+
+    UIStackView *_navigationBarStackView;
+    UIButton *_rendererButtton;
```
carolanitz: can we make these private properties instead of ivars ? also typo

Sources/VLCActionSheet.swift
Mikanbu: Yes, hooking `VLCActionSheet` with `PresentationTheme` is in the todo list~

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+ * Copyright © 2018 VLC authors and VideoLAN
+ * Copyright © 2018 Videolabs
+ *
+ * Authors: Soomin Lee <bubu@mikan.io>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+class VLCRendererDiscovererManager: NSObject {
+    @objc static let sharedInstance = VLCRendererDiscovererManager(presentingViewController: nil)
```
carolanitz: instead of having a sharedinstance we should use dependency injection. Also this is not how you create a sharedinstance you miss the dispatch_once but really this sharedinstance needs to go away

Sources/MediaViewController.swift
```diff
@@ -61,9 +64,21 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         setupCollectionView()
         setupSearchController()
         setupNavigationBar()
+        setupRendererDiscovererManager()
         _ = (MLMediaLibrary.sharedMediaLibrary() as! MLMediaLibrary).libraryDidAppear()
     }
 
+    public override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        let manager = VLCRendererDiscovererManager.sharedInstance
```
carolanitz: this code doesn't belong for me here. I would expect the manager to live inside the services

Sources/VLCMovieViewController.m
carolanitz: well just delete the second line

Sources/VLCMovieViewController.m
```diff
@@ -1649,4 +1673,36 @@ - (void)handleExternalScreenDidDisconnect:(NSNotification *)notification
...
+        _playingExternallyView.hidden = YES;
+    }
+    [_vpc mediaPlayerSetRenderer:renderer];
+}
+
+- (void)setupRendererDiscovererManager
+{
+    // Create a renderer button for VLCMovieViewController
+    _rendererButtton = [VLCRendererDiscovererManager.sharedInstance setupRendererButton];
+    [VLCRendererDiscovererManager.sharedInstance addSelectionHandlerWithSelectionHandler:^(VLCRendererItem * _Nonnull item) {
```
carolanitz: Does this result in a strong reference to the MovieViewController ? if so that would create a memoryleak

Sources/VLCPlaybackController.m
```diff
@@ -1290,4 +1292,14 @@ - (NSDictionary *)mediaOptionsDictionary
...
 }
 
+#pragma mark - Renderer
+
+- (void)mediaPlayerSetRenderer:(VLCRendererItem *)renderer
+{
+    if (_renderer != renderer) {
+        _renderer = renderer;
+    }
+    [_mediaPlayer setRendererItem:renderer];
```
carolanitz: into if ?

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+            if let rendererItem = item as? VLCRendererItem {
+                self?.setRendererItem(rendererItem: rendererItem)
+            }
+        }
+        return actionSheet
+    }()
+
+    @objc var presentingViewController: UIViewController?
+
+    @objc var rendererButttons: [UIButton] = [UIButton]()
```
carolanitz: ttt

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+    fileprivate init(presentingViewController: UIViewController?) {
+        self.presentingViewController = presentingViewController
+        super.init()
+    }
+
+    // Returns renderers of *all* discoverers
+    @objc func getAllRenderers() -> [VLCRendererItem] {
+        var renderers = [VLCRendererItem]()
+
+        for discoverer in discoverers {
```
carolanitz: use flatmap

Sources/VLCActionSheet.swift
Mikanbu: I'm not sure, I don't think that it's that of a fancy animation that it deserves to be global?

Sources/VLCActionSheet.swift
Mikanbu: This was mainly for a small animation purposes, I will add some comments

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+            renderers += discoverer.renderers
+        }
+        return renderers
+    }
+
+    fileprivate func isDuplicateDiscoverer(with description: VLCRendererDiscovererDescription) -> Bool {
+        for discoverer in discoverers {
```
carolanitz: that's inefficient. Why don't you use a set instead that is only used in the scope of start().

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+                    } else {
+                        print("Unable to start renderer discoverer with name: \(rendererDiscoverer.name)")
+                    }
+                } else {
+                    print("Unable to instanciate renderer discoverer with name: \(discoverer.name)")
+                }
+            }
+        }
+
+        return true
```
carolanitz: This returns true even if all VLCRendererDiscoverer.start() calls fail.

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+        return button
+    }
+}
+
+// MARK: VLCRendererDiscovererDelegate
+extension VLCRendererDiscovererManager: VLCRendererDiscovererDelegate {
+    func rendererDiscovererItemAdded(_ rendererDiscoverer: VLCRendererDiscoverer, item: VLCRendererItem) {
+        for button in rendererButttons {
+            if button.isHidden {
+                UIView.animate(withDuration: 0.1) {
```
carolanitz: this is a pretty short animation duration also I would expect a delegation pattern instead of the singleton owning UI that is shown elsewhere in the UI.

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+                    button.isHidden = false
+                }
+            }
+        }
+        actionSheet.collectionView.reloadData()
```
carolanitz: incremental updates would be better. Also: Doesn’t this triggers UI updates even when the view isn’t on screen?

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+                    }
+                    playbackController.mediaPlayerSetRenderer(nil)
+                }
+                // Reset buttons state
+                for button in rendererButttons {
+                    button.isSelected = false
+                }
+            }
+            actionSheet.collectionView.reloadData()
+            actionSheet.updateViewConstraints()
```
carolanitz: same as above

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+        }
+
+        // No more renderers to show
+        if getAllRenderers().isEmpty {
+            for button in rendererButttons {
+                button.isHidden = true
```
carolanitz: same as above

Sources/VLCActionSheet.swift
Mikanbu: Indeed, at first I wanted it to be an array of actions :D

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        setupView()
+        setupSeparator()
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        setupView()
```
carolanitz: if you have duplication in the inits use a common init also do you need both inits ? should one of them not be marked as not used?

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+}
+
+@objc protocol VLCActionSheetDataSource: class {
+    @objc func numberOfRows() -> Int
+    @objc func actionSheet(collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell
+}
+
+@objc protocol VLCActionSheetDelegate: class {
+    @objc optional func headerViewTitle() -> String?
+    @objc func itemAtIndexPath(_ indexPath: IndexPath) -> Any?
```
carolanitz: shouldn't the actionsheet be always the first param for a delegate call?

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+    @objc func actionSheet(collectionView: UICollectionView, didSelectItem item: Any, At indexPath: IndexPath)
```
carolanitz: Any is not great here can't we be more specific ?

Sources/VLCActionSheet.swift
```diff
@@ -0,0 +1,422 @@
...
+    @objc weak var delegate: VLCActionSheetDelegate?
+
+    var action: ((_ item: Any) -> Void)?
+
+    lazy var backgroundView: UIView = {
+        let backgroundView = UIView()
+        backgroundView.isHidden = true
+        backgroundView.autoresizingMask = [.flexibleHeight, .flexibleWidth]
+        backgroundView.backgroundColor = UIColor.black.withAlphaComponent(0.6)
+        backgroundView.isUserInteractionEnabled = true
```
carolanitz: isn't that the default ?

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+            return false
+        }
+        for discoverer in tmpDiscoverers {
+            if !isDuplicateDiscoverer(with: discoverer) {
+                if let rendererDiscoverer = VLCRendererDiscoverer(name: discoverer.name) {
+                    if rendererDiscoverer.start() {
+                        rendererDiscoverer.delegate = self
+                        discoverers.append(rendererDiscoverer)
+                    } else {
+                        print("Unable to start renderer discoverer with name: \(rendererDiscoverer.name)")
```
carolanitz: don't use prints but errors or asserts. When does this case happen ? Is there something the programmer did wrong or does the user need feedback that this failed ?

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+                    }
+                } else {
+                    print("Unable to instanciate renderer discoverer with name: \(discoverer.name)")
```
carolanitz: same as above

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+                }
+            }
+        }
+
+        return true
+    }
+
+    @objc func stop() {
+        for discoverer in discoverers {
```
carolanitz: map

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+
+                    if let movieViewController = presentingViewController as? VLCMovieViewController {
+                        movieViewController.setupCastWithCurrentRenderer()
+                    }
+                }
+            } else {
+                presentingViewController.present(actionSheet, animated: false, completion: nil)
+            }
+        } else {
+            print("VLCRendererDiscovererManager: Cannot display actionSheet, no viewController setted")
```
carolanitz: assert

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+                    actionSheet.collectionView.selectItem(at: indexPath, animated: false, scrollPosition: .centeredVertically)
+                    actionSheet(collectionView: actionSheet.collectionView, didSelectItem: rendererItem, At: indexPath)
+                    setRendererItem(rendererItem: rendererItem)
+
+                    if let movieViewController = presentingViewController as? VLCMovieViewController {
+                        movieViewController.setupCastWithCurrentRenderer()
+                    }
+                }
+            } else {
+                presentingViewController.present(actionSheet, animated: false, completion: nil)
```
carolanitz: why animated false ?

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+
+        if vpcRenderer != rendererItem {
+            VLCPlaybackController.sharedInstance().renderer = rendererItem
+            for button in rendererButtons {
+                button.isSelected = true
+            }
+        } else {
+            // Same renderer selected, removing selection
+            VLCPlaybackController.sharedInstance().renderer = nil
+            for button in rendererButtons {
```
carolanitz: you shouldn't need to iterate over all of them but be able to disable the one that was selected

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+        let vpcRenderer = VLCPlaybackController.sharedInstance().renderer
+
+        if vpcRenderer != rendererItem {
+            VLCPlaybackController.sharedInstance().renderer = rendererItem
+            for button in rendererButtons {
+                button.isSelected = true
+            }
+        } else {
+            // Same renderer selected, removing selection
+            VLCPlaybackController.sharedInstance().renderer = nil
```
carolanitz: Could we use dependency injection for the playbackcontroller as well here?

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+            for button in rendererButtons {
+                button.isSelected = false
+            }
+        }
+    }
+
+    @objc func addSelectionHandler(selectionHandler: ((_ rendererItem: VLCRendererItem) -> Void)?) {
+        actionSheet.setAction { [weak self] (item) in
+            if let rendererItem = item as? VLCRendererItem {
```
carolanitz: you forget to handle the else parts

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+        button.accessibilityHint = NSLocalizedString("BUTTON_RENDERER_HINT", comment: "")
+        rendererButtons.append(button)
+        return button
+    }
+}
+
+// MARK: VLCRendererDiscovererDelegate
+extension VLCRendererDiscovererManager: VLCRendererDiscovererDelegate {
+    func rendererDiscovererItemAdded(_ rendererDiscoverer: VLCRendererDiscoverer, item: VLCRendererItem) {
+        for button in rendererButtons {
```
carolanitz: map

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+    }
+}
+
+// MARK: VLCActionSheetDelegate
+extension VLCRendererDiscovererManager: VLCActionSheetDelegate {
+    func headerViewTitle() -> String? {
+        return NSLocalizedString("HEADER_TITLE_RENDERER", comment: "")
+    }
+
+    func itemAtIndexPath(_ indexPath: IndexPath) -> Any? {
```
carolanitz: this shouldn't return Any

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+        if indexPath.row < renderers.count {
+            return renderers[indexPath.row]
+        }
+        return nil
+    }
+
+    func actionSheet(collectionView: UICollectionView, didSelectItem item: Any, At indexPath: IndexPath) {
+        guard let renderer = item as? VLCRendererItem,
+            let cell = collectionView.cellForItem(at: indexPath) as? VLCActionSheetCell else {
+                return
```
carolanitz: no assert no error ?

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+// MARK: VLCActionSheetDataSource
+extension VLCRendererDiscovererManager: VLCActionSheetDataSource {
+    func numberOfRows() -> Int {
+        return getAllRenderers().count
+    }
+
+    func actionSheet(collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        guard let cell = collectionView.dequeueReusableCell(
+            withReuseIdentifier: VLCActionSheetCell.identifier, for: indexPath) as? VLCActionSheetCell else {
+                return UICollectionViewCell()
```
carolanitz: what happens actually when you just have UICollectionviewcell ? Is this still an acceptable state ?

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,257 @@
...
+        if indexPath.row < renderers.count {
+            let rendererName = renderers[indexPath.row].name
+            cell.name.text = rendererName
+
+            cell.icon.image = UIImage(named: "rendererGray")
+            if renderers[indexPath.row] == VLCPlaybackController.sharedInstance().renderer {
+                cell.icon.image = UIImage(named: "rendererFullOrange")
+            }
+        } else {
+            assertionFailure("VLCRendererDiscovererManager: cellForItemAt: IndexPath out of range")
```
carolanitz: 👍 but we still fall through here and return the cell in release is this okay ?

Sources/MediaViewController.swift
```diff
@@ -23,6 +23,8 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
     private var mediaDatasourceAndDelegate: MediaDataSourceAndDelegate?
     private var searchController: UISearchController?
     private let searchDataSource = VLCLibrarySearchDisplayDataSource()
+    private var rendererButton: UIButton
```
carolanitz: you don't need this

Sources/MediaViewController.swift
```diff
@@ -120,6 +135,11 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         collectionView?.backgroundView = collectionView?.numberOfItems(inSection: 0) == 0 ? emptyView : nil
     }
 
+    // MARK: Renderer
+    private func setupRendererButton() {
+        navigationItem.rightBarButtonItem = UIBarButtonItem(customView: rendererButton)
```
carolanitz: you just use rendererbutton directly from the manager

Sources/VLCActionSheet.swift
carolanitz: do you actually need a subclass here ? can't you just set that on a normal UICollectionViewFlowlayout ?

Sources/VLCActionSheet.swift
carolanitz: I would actually say that we should have a class for the different fonts and define them at a central point

Sources/VLCActionSheet.swift
carolanitz: also why 0.6 ?

Sources/VLCActionSheet.swift
Mikanbu: Maybe, at first I wanted to use a generic type and have the implementer use/inherit from it but for now I settled with Any.

I definitely want to improve `VLCActionSheet` later on!

Sources/VLCActionSheet.swift
Mikanbu: It is!

Sources/VLCRendererDiscovererManager.swift
Mikanbu: Because the animation is handled by `VLCActionSheet`.

Sources/MediaViewController.swift
Mikanbu: I'm not sure, I opted for a implementation of the renderer button by the VC that needs the renderer mostly because I want the possibility to the VC to customize their button if needed.
I'm not really happy by it because I  think the buttons should be contained in the manager but I didn't have a better idea. 😞 
For example: `MediaVC` and `VLCMovieVC` buttons have a different behaviour.

Sources/VLCRendererDiscovererManager.swift
Mikanbu: Would a `fatalError` be better?

Sources/VLCRendererDiscovererManager.swift
Mikanbu: Well, not sure how it should be handled, I can add an `assertionFailure` but won't it still go through in release?

Sources/VLCRendererDiscovererManager.swift
Mikanbu: I don't think we need the `dispatch_once`. 
DI, sure but wouldn't it better to wait till the coordinators are removed or we completely decide to use a coordinator pattern?

Sources/MediaViewController.swift
Mikanbu: You're right, I wasn't aware of the `VLCService` class.

Sources/VLCRendererDiscovererManager.swift
Mikanbu: Shouldn't it be a different PR to remove the `sharedInstance` of the `PlaybackController` then?

Sources/VLCRendererDiscovererManager.swift
Mikanbu: Not sure why a map here?

Sources/VLCActionSheet.swift
carolanitz: add a todo that these cells need to be dynamic in height in case someone uses big accessibilitytext

Sources/VLCActionSheet.swift
carolanitz: 0.3 is default and should be used could you otherwise share the animation with a gif / video to show what is actually going on here ?

Sources/VLCActionSheet.swift
carolanitz: the customization is too tiny to justify it's own class

Sources/VLCRendererDiscovererManager.swift
carolanitz: it has nothing to do with the coordinator. dependency injection would also be used if we turn the coordinators into viewcontrollers

Sources/VLCActionSheet.swift
Mikanbu: Well there is not much to see, `0.6` is only for the duration of the animation. I felt that 0.3 was way to fast.

Sources/VLCRendererDiscovererManager.swift
Mikanbu: Well a little bit no? Wouldn't it has to do with who instantiates the manager?

Sources/MediaViewController.swift
carolanitz: when do they need to customize the button ? if they don't need get rid of this and for different behaviors you have addaction target and remove action target

Sources/VLCActionSheet.swift
carolanitz: do it immediately otherwise you have more work later on

Sources/VLCRendererDiscovererManager.swift
Mikanbu: There is a private init on the manager 😢

carolanitz: I was wrong you were right 😁 regarding the private init. You've done it right. and regarding the who is initialising the manager it would be a veiwcontroller or whatever object takes the place of the coordinator

Sources/VLCRendererDiscovererManager.swift
carolanitz: yes good point

Sources/MediaViewController.swift
```diff
@@ -61,9 +64,21 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
...
 
+    public override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        let manager = services.rendererDiscovererManager
+        manager.start()
+        manager.presentingViewController = self
+    }
+
+    public override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
```
carolanitz: this can be deleted

Mikanbu: Indeed!

Sources/VLCActionSheet/VLCActionSheet.swift
```diff
@@ -0,0 +1,287 @@
...
+
+import Foundation
+import UIKit
+
+@objc protocol VLCActionSheetDataSource: class {
+    @objc func numberOfRows() -> Int
+    @objc func actionSheet(collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell
+}
+
+@objc protocol VLCActionSheetDelegate: class {
```
carolanitz: @objcMembers

Sources/VLCActionSheet/VLCActionSheet.swift
```diff
@@ -0,0 +1,287 @@
...
+ *
+ * Authors: Soomin Lee <bubu@mikan.io>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import Foundation
+import UIKit
+
+@objc protocol VLCActionSheetDataSource: class {
```
carolanitz: @objcmembers

Sources/VLCActionSheet/VLCActionSheet.swift
```diff
@@ -0,0 +1,287 @@
...
+        return collectionView
+    }()
+
+    lazy var cancelButton: UIButton = {
+        let cancelButton = UIButton()
+        cancelButton.setTitle(NSLocalizedString("CANCEL", comment: ""), for: .normal)
+        cancelButton.setTitleColor(.white, for: .normal)
+        cancelButton.titleLabel?.font = UIFont.systemFont(ofSize: 16)
+        cancelButton.addTarget(self, action: #selector(self.removeActionSheet), for: .touchDown)
+        cancelButton.backgroundColor = UIColor(red:1.00, green:0.59, blue:0.13, alpha:1.0)
```
carolanitz: this shouldn't be defined here where does this color come from ?

Mikanbu: Cancel button is not currently used, as you can see in `viewDidLoad` it's not initialized and a WIP.

Sources/VLCActionSheet/VLCActionSheet.swift
```diff
@@ -0,0 +1,287 @@
...
+        cancelButton.translatesAutoresizingMaskIntoConstraints = false
+        return cancelButton
+    }()
+
+    lazy var headerView: VLCActionSheetSectionHeader = {
+        let headerView = VLCActionSheetSectionHeader()
+        headerView.title.text = delegate?.headerViewTitle?() ?? "Default header title"
```
carolanitz: Default header title ?

Sources/VLCActionSheet/VLCActionSheet.swift
```diff
@@ -0,0 +1,287 @@
...
+        headerView.title.textColor = PresentationTheme.current.colors.cellTextColor
+        headerView.backgroundColor = PresentationTheme.current.colors.background
+        headerView.translatesAutoresizingMaskIntoConstraints = false
+        return headerView
+    }()
+
+    lazy var bottomBackgroundView: UIView = {
+        let bottomBackgroundView = UIView()
+        bottomBackgroundView.backgroundColor = UIColor(red:1.00, green:0.59, blue:0.13, alpha:1.0)
```
carolanitz: color

Sources/VLCActionSheet/VLCActionSheet.swift
```diff
@@ -0,0 +1,287 @@
...
+        let bottomBackgroundViewHeightConstraint = bottomBackgroundView.heightAnchor.constraint(equalToConstant: cellHeight)
+        return bottomBackgroundViewHeightConstraint
+    }()
+
+    @objc func removeActionSheet() {
+        UIView.transition(with: backgroundView, duration: 0.01, options: .transitionCrossDissolve, animations: {
+            [weak self] in
+            self?.backgroundView.isHidden = true
+            }, completion: { finished in
+                super.presentingViewController?.dismiss(animated: true, completion: nil)
```
carolanitz: super ?

Sources/VLCActionSheet/VLCActionSheet.swift
```diff
@@ -0,0 +1,287 @@
...
+@objc protocol VLCActionSheetDelegate: class {
+    @objc optional func headerViewTitle() -> String?
+    @objc func itemAtIndexPath(_ indexPath: IndexPath) -> Any?
+    @objc func actionSheet(collectionView: UICollectionView, didSelectItem item: Any, At indexPath: IndexPath)
+}
+
+// MARK: VLCActionSheet
+class VLCActionSheet: UIViewController {
+
+    private let cellHeight: CGFloat = 50
```
carolanitz: a static number is not good . maybe this can be done dynamically ?

Sources/VLCActionSheet/VLCActionSheet.swift
Mikanbu: I guess I could preconditionFailure here but it seemed a bit harsh.

Sources/VLCRendererDiscovererManager.swift
```diff
@@ -0,0 +1,284 @@
...
+        rendererButtons.append(button)
+        return button
+    }
+}
+
+// MARK: VLCRendererDiscovererDelegate
+extension VLCRendererDiscovererManager: VLCRendererDiscovererDelegate {
+    func rendererDiscovererItemAdded(_ rendererDiscoverer: VLCRendererDiscoverer, item: VLCRendererItem) {
+        for button in rendererButtons {
+            if button.isHidden {
```
carolanitz: you don't need the if here you van just set it to hidden no matter what

Sources/VLCActionSheet/VLCActionSheet.swift
Mikanbu: Indeed, totally missed it!

Mikanbu:
Merged by cb77dc7e74af1f6146ba8f3dee21dc0c8e92ef70.

PR #48, opened 2018-04-25
Fix set_changelog step for tvOS releases

dcordero:
### Checklist

### Description

As @carolanitz detected, the step `update_changelog` does not work as expected for tvOS releases, and it actually tries to set the changelog of the iOS application instead.

Since fastlane 2.90.0 🚀, there is new parameter `platform` available in the set_changelog action which allows fixing this issue.

This PR set this new parameter and, in addition to that, it converts the changelog file to a more standard and easier to parse format.


fastlane/Fastfile
```diff
@@ -95,12 +95,20 @@ end
...
   platform = options[:platform]
-  # Splits the News by -------- get out the top notes
-  changelog = File.read('../Docs/NEWS').split('-----------')[1].split('-----------').first
-  temp_changelog = changelog.split(platform)
-  temp_changelog = changelog.split('tvOS') if temp_changelog.count <= 1
-  changelog = temp_changelog[0..-2].join.strip
-  set_changelog(app_identifier: 'org.videolan.vlc-ios', changelog: changelog, username: '*', team_name: 'VideoLAN')
+  version = options[:version]
+
+  regex = /## #{platform} \[#{version}\](.*?)^##/m
```
carolanitz: that is sooo much better than what I had!!

carolanitz:
This looks great and I like the News format better than the old one

fastlane/Fastfile
dcordero: I cheated changing the format of the changelog, the previous format was much harder the parse 😉

dcordero:
@fkuehne done ✅. 

I changed them by - to make the document markdown compatible, but there is actually no need at all to make it a md file. I do agree that listing with with · looks much better than using minuses symbols.

Docs/NEWS
```diff
@@ -1,114 +1,124 @@
...
------------
+# Changelog
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## tvOS [Unreleased]
+
+## iOS [Unreleased]
```
dcordero: What do you think about these `unreleased` sections? 

The purpose of them is adding entries to the changelog progressively as soon as feature, fixes, etc... are implemented, getting at the end everything listed at the moment of releasing.

Does this workflow makes sense for vlc-ios?

carolanitz: I think they will probably disappear soon since I tend to just copy paste from the line below :D

carolanitz:
merged with 84b47471dd1c446071cc2d5c0785f16357735b85

PR #51, opened 2018-04-27
Lazy services

carolanitz:
### Checklist

### Description

I want to make sure that the Services are really just instantiated once and that was the best way I could think of :/ Is there a better way than this?


Mikanbu:
Merged.

PR #52, opened 2018-04-27
MediaDataSource: Added a Mediatype with Category and Subctegory to show the correct content on the subviews

carolanitz:
### Checklist

### Description

this adds the distinction between the different categories that we will introduce. I'm not sure if this is the best architecture but every MediaController has a Mediatype with Category and subcategory that can be set.  which also means that every object and controller that gets this injected can get notified about subcategory changes if needed. For example Sorting has different options based on the subcategory. Drag & Drop into albums should not be possible and so on


VLCMediaDataSource.swift
```diff
@@ -0,0 +1,275 @@
...
+            preconditionFailure("TODO")
+            return audioPlaylist
+        case .videoPlaylists:
+            preconditionFailure("TODO")
+            return videoPlaylist
+        case .allVideos:
+            return allVideos
+        }
+    }
+    @objc public func object(at index: Int, subcategory: VLCMediaSubcategory) -> NSManagedObject {
```
hectormatos2011: Add line break

VLCMediaDataSource.swift
```diff
@@ -0,0 +1,275 @@
...
+        super.init()
+        getAllVideos()
+        getAllAudio()
+    }
+
+    @objc public func numberOfFiles(subcategory: VLCMediaSubcategory) -> Int {
+        return array(for: subcategory).count
+    }
+
+    private func array(for subcategory: VLCMediaSubcategory ) -> [MLFile] {
```
hectormatos2011: extra space

carolanitz:
merged with b0f70796f70074132ad0a1191fdcf728d4c8ec60

PR #57, opened 2018-05-15
Fix empty view

mkchoi212:
### Checklist

### Description

simulator screen shot - iphone 6s - 2018-05-14 at 22 08 12

simulator screen shot - iphone 6s - 2018-05-14 at 22 03 01

- [x] Adapt empty view to theme
- [x] Adaptive layout
- [ ] Enable scrolling?

### DIFF
--- a/Sources/MediaViewController.swift
+++ b/Sources/MediaViewController.swift
@@ -37,7 +37,6 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         let name = String(describing: VLCEmptyLibraryView.self)
         let nib = Bundle.main.loadNibNamed(name, owner: self, options: nil)
         guard let emptyView = nib?.first as? VLCEmptyLibraryView else { fatalError("Can't find nib for \(name)") }
-        emptyView.frame = view.bounds
         return emptyView
     }()
 
@@ -60,7 +59,7 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
 
     @objc func reloadData() {
         collectionView?.reloadData()
-        displayEmptyViewIfNeeded()
+        updateUIForEmptyView()
     }
 
     @available(*, unavailable)
@@ -99,7 +98,7 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         super.viewDidAppear(animated)
         reloadData()
     }
-    
+
     func setupSearchController() {
         searchController = UISearchController(searchResultsController: nil)
         searchController?.searchResultsUpdater = self
@@ -112,20 +111,28 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
                 backgroundview.clipsToBounds = true
             }
         }
-        if #available(iOS 11.0, *) {
-            navigationItem.searchController = searchController
-        } else {
-            navigationItem.titleView = searchController?.searchBar
-            searchController?.hidesNavigationBarDuringPresentation = false
-        }
     }
-
+    
     func setupNavigationBar() {
         navigationItem.leftBarButtonItem = UIBarButtonItem(title: NSLocalizedString("SORT", comment: ""), style: .plain, target: self, action: #selector(sort))
     }
     
-    func displayEmptyViewIfNeeded() {
-        collectionView?.backgroundView = collectionView?.numberOfItems(inSection: 0) == 0 ? emptyView : nil
+    func updateUIForEmptyView() {
+        let isEmpty = collectionView?.numberOfItems(inSection: 0) == 0
+
+        if isEmpty {
+            collectionView?.backgroundView = emptyView
+            edgesForExtendedLayout = []
+        } else {
+            collectionView?.backgroundView = nil
+            edgesForExtendedLayout = .all
+        }
+        
+        if #available(iOS 11.0, *) {
+            navigationItem.searchController = isEmpty ? nil :searchController
+        } else {
+            navigationItem.titleView = isEmpty ? nil : searchController?.searchBar
+        }
     }
 
     @objc func sort() {
--- a/Sources/MediaViewController.swift
+++ b/Sources/MediaViewController.swift
@@ -121,6 +121,7 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         let isEmpty = collectionView?.numberOfItems(inSection: 0) == 0
 
         if isEmpty {
+            collectionView?.setContentOffset(.zero, animated: false)
             collectionView?.backgroundView = emptyView
             edgesForExtendedLayout = []
         } else {
--- a/Sources/VLCEmptyLibraryView.h
+++ b/Sources/VLCEmptyLibraryView.h
@@ -16,7 +16,6 @@
 
 @property (nonatomic, strong) IBOutlet UILabel *emptyLibraryLabel;
 @property (nonatomic, strong) IBOutlet UILabel *emptyLibraryLongDescriptionLabel;
-@property (nonatomic, strong) IBOutlet UIActivityIndicatorView *activityIndicator;
 @property (nonatomic, strong) IBOutlet UIButton *learnMoreButton;
 
 - (IBAction)learnMore:(id)sender;
--- a/Sources/VLCEmptyLibraryView.m
+++ b/Sources/VLCEmptyLibraryView.m
@@ -23,11 +23,10 @@
     _emptyLibraryLongDescriptionLabel.text = NSLocalizedString(@"EMPTY_LIBRARY_LONG", nil);
     [_learnMoreButton setTitle:NSLocalizedString(@"BUTTON_LEARN_MORE", nil) forState:UIControlStateNormal];
     
-    NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
-    [notificationCenter addObserver:self
-                           selector:@selector(themeDidChange)
-                               name:kVLCThemeDidChangeNotification
-                             object:nil];
+    [[NSNotificationCenter defaultCenter] addObserver:self
+                                             selector:@selector(themeDidChange)
+                                                 name:kVLCThemeDidChangeNotification
+                                               object:nil];
     [self themeDidChange];
     [super awakeFromNib];
 }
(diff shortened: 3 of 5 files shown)


carolanitz:
I will need to checkout the code and test but the screenshots look great. I think scrolling should still be enabled because I'm not sure if you are able to display the entire content on an iPhone 4s especially when we have the top bar in there as well for the different categories.

Sources/VLCEmptyLibraryView.h
```diff
@@ -16,7 +16,6 @@
 
 @property (nonatomic, strong) IBOutlet UILabel *emptyLibraryLabel;
 @property (nonatomic, strong) IBOutlet UILabel *emptyLibraryLongDescriptionLabel;
-@property (nonatomic, strong) IBOutlet UIActivityIndicatorView *activityIndicator;
```
carolanitz: 👍

Sources/MediaViewController.swift
```diff
@@ -74,10 +74,16 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         setupSearchController()
         setupNavigationBar()
         _ = (MLMediaLibrary.sharedMediaLibrary() as! MLMediaLibrary).libraryDidAppear()
+        themeDidChange()
     }
 
     @objc func themeDidChange() {
         collectionView?.backgroundColor = PresentationTheme.current.colors.background
+        guard let backgroundView = collectionView?.backgroundView as? VLCEmptyLibraryView else {
```
carolanitz: Since you're changing the emptyview it would be better to move this code into the VLCEmptyLibraryView. The MedialibraryViewcontroller should not know or modify the properties of the Emptyview. 
The reason why you do this. is because you might want to reuse the Emptyview somewhere else and then the controller who is using emptyview would have to change these properties as well -> duplication of code. If you do it in the emptyview itself this code is easier to find and second the class is easier to be reused. 

Observe the notification there and add a themeDidChange method in VLCEmptyLibraryView to update these properties.

Sources/MediaViewController.swift
```diff
@@ -93,6 +99,7 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
             collectionView?.dragDelegate = dragAndDropManager
             collectionView?.dropDelegate = dragAndDropManager
         }
+        reloadData()
```
carolanitz: why do you need this ?

Sources/MediaViewController.swift
```diff
@@ -74,10 +74,16 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         setupSearchController()
         setupNavigationBar()
         _ = (MLMediaLibrary.sharedMediaLibrary() as! MLMediaLibrary).libraryDidAppear()
+        themeDidChange()
```
carolanitz: I don't think this is needed anymore when you move the code

carolanitz:
@Mikanbu scaling it and disabling scrolling will be more code and state then simply adding the view and let it still be scrollable

Mikanbu:
Sure, then I think we should at least do something for the navigationBar, it feels wrong with an emptyView.

carolanitz:
@Mikanbu I think the navigationbar can stay but I would remove the search as well

Mikanbu:
@carolanitz Yes, sorry I should have been more clear, I meant the animation of the navigationBar seems wrong with the empty view.

carolanitz:
btw here in the image the notch is not respected and I was rotating the device. It's great that the image and text arrange correctly when you rotate but after rotating to landscape the image and text should be more centered :) 
img_1975

carolanitz:
okay to recap the changes that we need are:

- [ ] make it scrollable
- [ ] hide and show the searchbar depending on if there is content or not
- [ ] center the image and text more in landscape
- [ ] move the theme did change code into emptylibraryview

carolanitz:
@Mikanbu can you be more specific ? what part of the animation feels wrong and how should it be ?

Sources/MediaViewController.swift
```diff
@@ -100,6 +100,22 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
         reloadData()
     }
     
+    func showSearchController(_ show: Bool) {
+        if show {
```
carolanitz: It's very hard to read, can you change that to: 
``` 
 if #available(iOS 11.0, *) {
                navigationItem.searchController = show  ? searchController : nil
           } else {
                navigationItem.titleView = show  ? searchController?.searchBar : nil
            } 
```

Sources/MediaViewController.swift
```diff
@@ -112,20 +128,21 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
                 backgroundview.clipsToBounds = true
             }
         }
-        if #available(iOS 11.0, *) {
-            navigationItem.searchController = searchController
-        } else {
-            navigationItem.titleView = searchController?.searchBar
-            searchController?.hidesNavigationBarDuringPresentation = false
-        }
+        showSearchController(true)
```
carolanitz: Actually, don't have this state with true. It should be determined by the method itself and call it showSearchControllerIfNeeded. In there you will check the collectionview.numberOfSections

Sources/MediaViewController.swift
```diff
@@ -112,20 +128,21 @@ public class VLCMediaViewController: UICollectionViewController, UISearchResults
...
     }
-
+    
     func setupNavigationBar() {
         navigationItem.leftBarButtonItem = UIBarButtonItem(title: NSLocalizedString("SORT", comment: ""), style: .plain, target: self, action: #selector(sort))
     }
     
     func displayEmptyViewIfNeeded() {
-        collectionView?.backgroundView = collectionView?.numberOfItems(inSection: 0) == 0 ? emptyView : nil
+        if collectionView?.numberOfItems(inSection: 0) == 0 {
```
carolanitz: hmm thinking about it we can probably move the searchbar code in here as well and adjust the methodname to reflect that it updates the UI for empty state or not

carolanitz: yeah so just delete showSearchController  and change this Method to this : 
```
Bool isEmpty  =  collectionView?.numberOfItems(inSection: 0) == 0
collectionView?.backgroundView = isEmpty ? emptyView : nil
if #available(iOS 11.0, *) {
  navigationItem.searchController =  isEmpty ? nil : searchController
} else {
  navigationItem.titleView =  isEmpty ? nil : searchController?.searchBar
} 
```

mkchoi212:
According to Apple,
> The view (if any) in this property is **positioned underneath all of the other content and sized automatically to fill the entire bounds of the collection view.** The background view **does not scroll with the collection view’s other content.**
> https://developer.apple.com/documentation/uikit/uicollectionview/1618074-backgroundview

Maybe add the view as a cell/footer of the collectionView? I feel like this would lead to other edge cases and make things complicated :p

Sources/VLCEmptyLibraryView.m
```diff
@@ -21,6 +22,13 @@ - (void)awakeFromNib
     _emptyLibraryLabel.text = NSLocalizedString(@"EMPTY_LIBRARY", nil);
     _emptyLibraryLongDescriptionLabel.text = NSLocalizedString(@"EMPTY_LIBRARY_LONG", nil);
     [_learnMoreButton setTitle:NSLocalizedString(@"BUTTON_LEARN_MORE", nil) forState:UIControlStateNormal];
+    
+    NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
+    [notificationCenter addObserver:self
```
Mikanbu: I think we don't need to cache the default center here since it's only used once!
So I would suggest to directly use `[[NSNotificationCenter defaultCenter] addObserver:...]`

carolanitz:
I think this will be on my plate since I need to take a look at making the empty view scroll with the rest when I merge this. But yeah that doesn't look good

mkchoi212:
simulator screen shot - iphone 5s - 2018-05-21 at 22 56 10

The weird layout problem has been fixed... partially
There is one minor bug where the position of the view changes after the view is scrolled.

mkchoi212:
Problem has been solved by collapsing the navigation bar as suggested by @Mikanbu!

Mikanbu:
Small thought, shouldn't `Sort` be disable/hidden ? 🤔

carolanitz:
I think this can be ignored in this PR case since this will change with my subcategory PR. I think I need to adjust the search with that one as well...

Sources/VLCTabBarCoordinator.swift
```diff
@@ -82,7 +82,11 @@ class VLCTabbarCooordinator: NSObject, VLCMediaViewControllerDelegate {
         settingsVC.tabBarItem.accessibilityIdentifier = VLCAccessibilityIdentifier.settings
 
         let controllers = [audioVC, serverVC, videoVC, settingsVC]
-        tabBarController.viewControllers = controllers.map { UINavigationController(rootViewController: $0)}
+        tabBarController.viewControllers = controllers.map {
+            let navigationController = UINavigationController(rootViewController: $0)
+            navigationController.navigationBar.isTranslucent = true
```
carolanitz: how did that all of a sudden sneak in there ?!

carolanitz:
Actually this doesn't lint and please remove the translucent code block. I don't think that's necessary  ```tabBarController.viewControllers = controllers.map {
+            let navigationController = UINavigationController(rootViewController: $0)
+            navigationController.navigationBar.isTranslucent = true
 +            return navigationController```

Sources/VLCTabBarCoordinator.swift
mkchoi212: That's how I prevented the empty view from being underneath the navigation bar. Is there another way to enforce that?

carolanitz: I actually don't think you should set the emptyview as backgroundview but add it as subview of the collectionview and then this shouldn't happen and you don't need to set the contentoffset

mkchoi212: You were right! Turns out I don't need this bit. But I couldn't get it to work by adding it as a subview. I used `view.bounds` instead of `collectionView.bounds` when adding the view :(

carolanitz:
merged with 6397c5f66075652ba253864115c8a674422e42f2

PR #59, opened 2018-05-17
Subcategory collectionview

carolanitz:
### Checklist

### Description

This is totally breaking tests, linters and everything but I just wanted to share what I'm currently working on :) No review yet necessary. I'll update as I got but this just also helps future discussions okay byyeeee
ezgif com-video-to-gif 2


carolanitz:
only things missing is it needs to update to the theme and there is a weird offset on iOS 10

Resources/en.lproj/Localizable.strings
```diff
@@ -51,9 +51,7 @@
 "CONTINUE_PLAYBACK" = "Continue playback?";
 "CONTINUE_PLAYBACK_LONG" = "Would you like to continue playback of \"%@\" where you left off?";
 
-"BUTTON_ALL" = "Select All";
```
carolanitz: These changes shouldn't be in here because we still nee those strings

Sources/ButtonBarView.swift
```diff
@@ -0,0 +1,129 @@
+//
+//  ButtonBarView.swift
+//  VLC-iOS
+//
+//  Created by Carola Nitz on 6/1/18.
+//  Copyright © 2018 VideoLAN. All rights reserved.
```
carolanitz: header

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,647 @@
...
+            NSLayoutConstraint.activate([
+                buttonBarView.topAnchor.constraint(equalTo: topLayoutGuide.bottomAnchor),
+                buttonBarView.rightAnchor.constraint(equalTo: view.rightAnchor),
+                buttonBarView.leftAnchor.constraint(equalTo: view.leftAnchor),
+                buttonBarView.heightAnchor.constraint(equalToConstant: 35)
+                ])
+            NSLayoutConstraint.activate([
+                containerView.topAnchor.constraint(equalTo: buttonBarView.bottomAnchor),
+                containerView.rightAnchor.constraint(equalTo: view.rightAnchor),
+                containerView.leftAnchor.constraint(equalTo: view.leftAnchor),
```
carolanitz: this is not great that the cell lays out it superclasses view

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,647 @@
...
+    open override func viewDidLoad() {
+        super.viewDidLoad()
+
+        let buttonBarViewAux = buttonBarView ?? {
+            let flowLayout = UICollectionViewFlowLayout()
+            flowLayout.scrollDirection = .horizontal
+            let buttonBar = ButtonBarView(frame: .zero, collectionViewLayout: flowLayout)
+            buttonBar.backgroundColor = .white
+            buttonBar.selectedBar.backgroundColor = PresentationTheme.current.colors.orangeUI
+            buttonBar.scrollsToTop = false
```
carolanitz: move the color code and all that into the ButtonBarView itself

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,647 @@
...
+        fatalError("You must override this method to set up ButtonBarView cell accordingly")
+    }
+
+    private func calculateWidths() -> [CGFloat] {
+        let flowLayout = buttonBarView.collectionViewLayout as! UICollectionViewFlowLayout // swiftlint:disable:this force_cast
+        let numberOfCells = viewControllers.count
+
+        var minimumCellWidths = [CGFloat]()
+        var collectionViewContentWidth: CGFloat = 0
+        let indicatorWidth:CGFloat = 70.0
```
carolanitz: magic constant should be at the top

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,647 @@
...
+        }
+        return .none
+    }
+
+    override open func viewDidLoad() {
+        super.viewDidLoad()
+        let containerViewAux = containerView ?? {
+            return UIScrollView(frame: CGRect(x: 0, y: 0, width: view.bounds.width, height: view.bounds.height))
+            }()
+        containerView = containerViewAux
```
carolanitz: somehow this style doesn't mix well with the rest of the swift code. Can you adapt that to fit in more with the rest?

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,647 @@
...
+}
+
+public protocol PagerTabStripDataSource: class {
+
+    func viewControllers(for pagerTabStripController: PagerTabStripViewController) -> [UIViewController]
+}
+
+// MARK: PagerTabStripViewController
+
+open class PagerTabStripViewController: UIViewController, UIScrollViewDelegate {
```
carolanitz: put this into it's own file

Sources/VLCMediaDataSource.swift
```diff
@@ -42,58 +42,79 @@ struct VLCMediaType {
...
+        case .audioPlaylists:
+            return IndicatorInfo(title: NSLocalizedString("AUDIO_PLAYLISTS", comment: ""))
+        case .videoPlaylists:
+            return IndicatorInfo(title: NSLocalizedString("VIDEO_PLAYLISTS", comment: ""))
+        case .allVideos:
+            return IndicatorInfo(title: NSLocalizedString("VIDEOS", comment: ""))
+        }
+
+    }
+
```
carolanitz: double space

Sources/VLCMediaDataSource.swift
```diff
@@ -139,30 +160,88 @@ struct VLCMediaType {
...
+        movies = foundVideos.filter{ $0.isMovie() }
+    }
+//    private func tracksFromAudio() {
+//        if tracks != foundAudio {
+//            tracks = foundAudio
+//            NotificationCenter.default.post(name: .VLCTracksDidChangeNotification, object: tracks)
+//        }
+//    }
+//
+//    private func allVideosFromVideos() {
```
carolanitz: if you don't need that delete that

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,647 @@
...
+ * $Id$
+ *
+ * Authors: Carola Nitz <caro # videolan.org>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import Foundation
+
+class VLCLabelCell: UICollectionViewCell {
```
carolanitz: now that you removed the image do you really need a subclass?

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,647 @@
...
+                containerView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+                ]
+            )
+        }
+
+        buttonBarView.delegate = self
+        buttonBarView.dataSource = self
+
+        // register button bar item cell
+        buttonBarView.register(UINib(nibName: "VLCLabelCell", bundle: .main), forCellWithReuseIdentifier:"Cell")
```
carolanitz: should also go into the buttonbar

Sources/BaseButtonBarPagerTabStripViewController.swift
carolanitz: yes because there is no UIlabel duhhh this is not a UITableViewCell

Sources/PagerStripViewController.swift
```diff
@@ -0,0 +1,356 @@
...
+        viewControllers = dataSource.viewControllers(for: self)
+
+        guard !viewControllers.isEmpty else {
+            fatalError("viewControllers(for:) should provide at least one child view controller")
+        }
+        viewControllers.forEach { if !($0 is IndicatorInfoProvider) { fatalError("Every view controller provided by PagerTabStripDataSource's viewControllers(for:) method must conform to IndicatorInfoProvider") }}
+
+    }
+
+    private var pagerTabStripChildViewControllersForScrolling: [UIViewController]?
```
Mikanbu: I think properties, even private, should be declared at the top. Is there a particular reason why you prefer them at the bottom?

Mikanbu:
It seems that the navigation bar iOS 11 style is disabled? 🤔

Mikanbu:
Looking at [figma](https://www.figma.com/proto/ByhPBHr4gTmUwo4bsVcanWIN/VLC180510?scaling=contain&redirected=1&node-id=0%3A193), I think the tabs/sub-categories should be a bit more "spaced"?

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,290 @@
...
+    }
+}
+
+public enum PagerScroll {
+    case no
+    case yes
+    case scrollOnlyIfOutOfScreen
+}
+
+open class BaseButtonBarPagerTabStripViewController<ButtonBarCellType: UICollectionViewCell>: PagerTabStripViewController, PagerTabStripDataSource, PagerTabStripIsProgressiveDelegate, UICollectionViewDelegate, UICollectionViewDataSource {
```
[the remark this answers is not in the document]
carolanitz: yes I'll remove the open. I actually just looked up if prefixes are still needed and for swift it's not. I'm actually wondering if we should rename all Swift classes to get rid of the VLC prefix altogether ? obviously in a separate PR

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,290 @@
...
+
+    @objc open func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAtIndexPath indexPath: IndexPath) -> CGSize {
+        guard let cellWidthValue = cachedCellWidths?[indexPath.row] else {
+            fatalError("cachedCellWidths for \(indexPath.row) must not be nil")
+        }
+        return CGSize(width: cellWidthValue, height: collectionView.frame.size.height)
+    }
+
+    open func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
+        guard indexPath.item != currentIndex else { return }
```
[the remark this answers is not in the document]
carolanitz: no. I'm not sure if you checked what this code actually does, but in case you didn't It checks if the currently selected index is the same as the one we just tapped and in that case there is nothing to do here because the view is already in the correct state

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,290 @@
...
+            for minimumCellWidthValue in minimumCellWidths {
+                let cellWidth = (minimumCellWidthValue > generalMinimumCellWidth) ? minimumCellWidthValue : generalMinimumCellWidth
+                stretchedCellWidths.append(cellWidth)
+            }
+
+            return stretchedCellWidths
+        }
+    }
+
+    private var shouldUpdateButtonBarView = true
```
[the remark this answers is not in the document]
carolanitz: This is mostly from the copying out the parts we needed out of the other lib

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,290 @@
...
+ *****************************************************************************/
+
+import Foundation
+
+class VLCLabelCell: UICollectionViewCell {
+
+    @IBOutlet weak var iconLabel: UILabel!
+
+}
+public enum SwipeDirection {
```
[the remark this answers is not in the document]
carolanitz: yes we do for these actually

carolanitz:
>It seems that the navigation bar iOS 11 style is disabled? 🤔

yes that's wanted otherwise it would jump when you swipe from empty to non emptyviewcontrollers 

> Looking at figma, I think the tabs/sub-categories should be a bit more "spaced"?

yes the font was one size too big! Well spotted

Sources/MediaSubcategoryViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+            oldCell?.iconLabel.textColor = PresentationTheme.current.colors.cellDetailTextColor
+            newCell?.iconLabel.textColor = PresentationTheme.current.colors.orangeUI
+        }
+        if #available(iOS 11.0, *) {
+            navigationController?.navigationBar.prefersLargeTitles = false
+        }
+        super.viewDidLoad()
+    }
+
+    // MARK: - PagerTabStripDataSource
```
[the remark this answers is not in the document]
carolanitz: I get errors "@objc is not supported within extensions of generic classes or classes that inherit from generic classes"

Mikanbu:
Thank you for adjusting the view!

However, I'm really sorry but I'm going to be a bit annoying on this. 😅 
I tried to make an image in order to illustrate my thoughts on the spacing:

https://gyazo.com/b0e44709903e70f68e247076b56b0434

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,294 @@
...
+ *
+ * Authors: Carola Nitz <caro # videolan.org>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import Foundation
+
+class VLCLabelCell: UICollectionViewCell {
+    static var cellIdentifier = "VLCLabelCell"
```
Mikanbu: I think a `static let` might be better for an identifier.

Sources/BaseButtonBarPagerTabStripViewController.swift
```diff
@@ -0,0 +1,290 @@
...
+    required public init?(coder aDecoder: NSCoder) {
+        fatalError()
+    }
+
+    open override func viewDidLoad() {
+        super.viewDidLoad()
+
+        let flowLayout = UICollectionViewFlowLayout()
+        flowLayout.scrollDirection = .horizontal
+        buttonBarView = ButtonBarView(frame: .zero, collectionViewLayout: flowLayout)
```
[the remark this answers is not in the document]
Mikanbu: You're right, it's not really the laziness that mattered the most to me here!
What I wanted us to consider is a "closure initialization"(not sure about the name 🤔) so that all(or most of) the `buttonBarView` related initialization is done next to its declaration.

One of the big benefits of it would be to have a more "segregated/readable" code by:

- "Containing" the initialization at the top
- Reducing `viewDidLoad` size

Small example in `VLCActionSheet`

Sources/BaseButtonBarPagerTabStripViewController.swift
carolanitz: very true I think I have this somewhere else as well so I will change that

carolanitz:
Really like the points you brought up and I will adjust both the spacing and the init.

carolanitz:
closed with a massive commit dump starting at a89d9f9be68db108ed2df637ccc32c4d86f5b16d ending in 46ea41379e604a667e78a784b97734d11ea2567d

PR #61, opened 2018-05-18
Fix warnings

carolanitz:
### Checklist

### Description

Warnings count is still 52 but it's going down!
This updates the project settings to include more warnings and it also fixed a bunch of implicit uses of self. I personally find the self-> code quite ugly  but the only change to fix that would be to switch to properties which I wanted to put up for discussion at this point


carolanitz:
merged with 
commit 850a3836ab16dde29439f9c99a24e83774afde21
commit ebc04867b37b440941491ce9998b80f0ca6fd383
and 
commit bbfa41ff00188f094f41b2cf1efec2fce8257c73

PR #75, opened 2018-06-05
VLCTabBarCoordinator: Change tab order

Mikanbu:
Move the Video tab to first posisiton

### Checklist

### Description

Move the video tab to first position according to the new design.


carolanitz:
yeah we can merge that but it's also part of the subcategory :) What would you prefer?

Mikanbu:
Indeed, I saw that only after this PR 😅 
I do not have a particular preference, even though, I think it shouldn't be in the [subcategory collectionview](https://github.com/videolan/vlc-ios/pull/59) PR(I didn't even think of checking there!).
So we could either cherry-pick your commit from your PR or merge this.
At the end of the day, I just want the tabs to be in the right order! 🎉

Mikanbu:
Merged with b7514ad83018ca7c55c297171861b87484b5d160.

PR #76, opened 2018-06-08
Settings view

mkchoi212:
### Checklist

### Description

Settings view has been updated to match the mockups on Figma (almost)

There are still couple of things that need to be fixed but thought I would open this PR to share with you guys what I have so far 😄


Sources/VLCSettingsController.m
```diff
@@ -125,12 +138,37 @@ - (void)updateUIAndCoreSpotlightForPasscodeSetting:(BOOL)passcodeOn
...
+    UILabel *headerLabel = [[UILabel alloc] init];
+    
+    headerLabel.font = [UIFont systemFontOfSize:23 weight:UIFontWeightSemibold];
+    headerLabel.text = [settingsViewController.settingsReader titleForSection:section];
+    headerLabel.textColor = PresentationTheme.current.colors.cellTextColor;
+    headerLabel.translatesAutoresizingMaskIntoConstraints = false;
+    [header addSubview:headerLabel];
+    
+    UILayoutGuide *margins = [header layoutMarginsGuide];
+    [[headerLabel.leadingAnchor constraintEqualToAnchor:margins.leadingAnchor constant:8] setActive:true];
```
Mikanbu: I'm aware that this is a WIP so I'm just going to leave a single comment! 😄 
What would you think about using `[NSLayoutConstraint activateConstraints:@[];`?

mkchoi212: I didn't know that was a thing! Definitely makes it look cleaner :D

mkchoi212: If you don't mind, please put more comments on stuff! I'm kinda stuck on most things anyways 🤦‍♂️

Mikanbu: Yes, and if we trust the [documentation](https://developer.apple.com/documentation/uikit/nslayoutconstraint/1526955-activateconstraints), it's even ~sometimes~ faster! 😄

Resources/Settings.bundle/Root.inApp.plist
```diff
@@ -33,6 +13,14 @@
 			<string>SETTINGS_DARKTHEME</string>
 			<key>Type</key>
 			<string>PSToggleSwitchSpecifier</string>
+			<key>IASKSubtitle</key>
+			<string>Change app colours</string>
```
mkchoi212: How do I go about adding a new string that needs to be localized??

Resources/Settings.bundle/Root.inApp.plist
```diff
@@ -43,6 +31,8 @@
 			<string>SETTINGS_PASSCODE_LOCK</string>
 			<key>Type</key>
 			<string>PSToggleSwitchSpecifier</string>
+			<key>IASKSubtitle</key>
+			<string>Secure access to your content</string>
```
mkchoi212: Same thing here. Needs localization

Sources/VLCSettingsController.m
```diff
@@ -171,4 +206,30 @@ - (void)updateForPasscode:(NSString *)passcode
...
+        cell = [[UITableViewCell alloc] initWithStyle:UITableViewCellStyleSubtitle reuseIdentifier:identifier];
+        cell.accessoryType = [identifier isEqualToString:kIASKPSMultiValueSpecifier] ? UITableViewCellAccessoryDisclosureIndicator : UITableViewCellAccessoryNone;
+    }
+    else {
+        cell = [[UITableViewCell alloc] initWithStyle:UITableViewCellStyleSubtitle reuseIdentifier:identifier];
+        if ([identifier isEqualToString:kIASKOpenURLSpecifier]) {
+            cell.accessoryType = UITableViewCellAccessoryDisclosureIndicator;
+        }
+    }
+    IASK_IF_PRE_IOS6(cell.textLabel.minimumFontSize = kIASKMinimumFontSize;
```
Mikanbu: I think we can remove this bit of code since we don't support pre-iOS 6 anymore?

Sources/VLCSettingsController.m
```diff
@@ -125,12 +138,38 @@ - (void)updateUIAndCoreSpotlightForPasscodeSetting:(BOOL)passcodeOn
...
-- (void)settingsViewController:(IASKAppSettingsViewController*)sender buttonTappedForSpecifier:(IASKSpecifier*)specifier
+- (IBAction)pushAboutViewController:(id)sender
 {
     VLCAboutViewController *aboutVC = [[VLCAboutViewController alloc] init];
     [self.navigationController pushViewController:aboutVC animated:YES];
 }
 
+- (UIView *)settingsViewController:(id<IASKViewController>)settingsViewController tableView:(UITableView *)tableView viewForHeaderForSection:(NSInteger)section
+{
+    UIView *header = [[UIView alloc] initWithFrame:CGRectMake(20, 0, tableView.bounds.size.width, 64)];
```
Mikanbu: I think we should consider using a macro at the top for the default values.
For example something like:
`#define SETTINGS_HEADER_HEIGHT 64`

We could also use a `const` for this if you prefer.

Sources/VLCSettingsController.m
```diff
@@ -35,16 +42,22 @@ - (instancetype)initWithStyle:(UITableViewStyle)style
...
 - (void)viewDidLoad
 {
+    [super viewDidLoad];
+
     self.modalPresentationStyle = UIModalPresentationFormSheet;
     self.delegate = self;
     self.showDoneButton = NO;
     self.showCreditsFooter = NO;
+    
+    self.navigationItem.leftBarButtonItem = [[UIBarButtonItem alloc] initWithTitle:NSLocalizedString(@"About", nil) style:UIBarButtonItemStylePlain target:self action:@selector(pushAboutViewController:)];
```
Mikanbu: Are you sure about the `About` ~☜(゚ヮ゚☜)~ string the right localization identifier?

Sources/VLCSettingsController.m
```diff
@@ -125,12 +138,38 @@ - (void)updateUIAndCoreSpotlightForPasscodeSetting:(BOOL)passcodeOn
...
+                                              [headerLabel.trailingAnchor constraintEqualToAnchor:margins.trailingAnchor],
+                                              [headerLabel.topAnchor constraintEqualToAnchor:margins.topAnchor],
+                                              [headerLabel.bottomAnchor constraintEqualToAnchor:margins.bottomAnchor]
+                                              ]];
+    return header;
+}
+
+- (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    return 64.0;
```
Mikanbu: What would you think about having a dynamic height?_(mostly for accessibility reasons)_
Like for example what @carolanitz  did in bf01cc4477fe420f36371605acb48d6c4a5a06fe

Resources/Settings.bundle/Root.inApp.plist
fkuehne: You need to add constants and their plain text counter-part to the en.lproj's "Root.strings" within the Settings.bundle.

Resources/Settings.bundle/Root.inApp.plist
fkuehne: See above :)

Sources/VLCSettingsController.m
```diff
@@ -35,16 +42,22 @@ - (instancetype)initWithStyle:(UITableViewStyle)style
 
 - (void)viewDidLoad
 {
+    [super viewDidLoad];
+
```
fkuehne: 👍

Sources/VLCSettingsController.m
fkuehne: Yes, that's a good idea and matches what we have in other classes.

Sources/VLCSettingsController.m
```diff
@@ -125,12 +138,38 @@ - (void)updateUIAndCoreSpotlightForPasscodeSetting:(BOOL)passcodeOn
...
     VLCAboutViewController *aboutVC = [[VLCAboutViewController alloc] init];
     [self.navigationController pushViewController:aboutVC animated:YES];
 }
 
+- (UIView *)settingsViewController:(id<IASKViewController>)settingsViewController tableView:(UITableView *)tableView viewForHeaderForSection:(NSInteger)section
+{
+    UIView *header = [[UIView alloc] initWithFrame:CGRectMake(20, 0, tableView.bounds.size.width, 64)];
+    UILabel *headerLabel = [[UILabel alloc] init];
+    
+    headerLabel.font = [UIFont systemFontOfSize:23 weight:UIFontWeightSemibold];
```
fkuehne: Can we generalize that font configuration similar to the way we handle colors in a global sense?

Sources/VLCSettingsController.m
fkuehne: oh yes!

Sources/VLCSettingsController.m
mkchoi212: That's a good idea but I'm not really sure how to do make a cell have a minimum height of 64 px - as shown in the mockups - when I'm using the default `UITableViewCellStyleSubtitle` for example. Is making my own custom cell the only option?

mkchoi212:
With https://github.com/videolan/vlc-ios/pull/76/commits/672bda142d81628a8d01af376e3ecef25800e35a, the privacy cell is in the same section as the theme cell as shown in the mockup!

simulator screen shot - iphone x - 2018-06-11 at 10 31 32

mkchoi212:
Final view has...
- Section header with separator
- Subtitle style for applicable cells
- Cell separator turned off
- Minimum cell height of `64px`
- New labeling on headers
- About button on top left corner

simulator screen shot - iphone x - 2018-06-13 at 13 06 08

Mikanbu:
Sorry, small nitpicking again 😄 
When in dark mode, the `thumbTintColor` of the switches should be changed as well like:
 https://gyazo.com/ea524b9883585fb97d504c5de34705be

mkchoi212:
I noticed that UISwitch in off-state *(both dark and light mode)* is not in the Figma mockups. So I took the liberty and made one on the fly 😄 

switch

SharedSources/PresentationTheme.swift
```diff
@@ -31,9 +39,8 @@ extension Notification.Name {
     public let lightTextColor: UIColor
     public let sectionHeaderTextColor: UIColor
     public let sectionHeaderTintColor: UIColor
-    public let settingsBackground: UIColor
-    public let settingsCellBackground: UIColor
     public let settingsSeparatorColor: UIColor
+    public let switchBackgroundColor: UIColor
```
carolanitz: I think we also need that actually in the local network screen

Sources/AppearanceManager.swift
```diff
@@ -37,9 +37,14 @@ class AppearanceManager: NSObject {
         // For the edit selection indicators
         UITableView.appearance().tintColor = theme.colors.orangeUI
         UISegmentedControl.appearance().tintColor = theme.colors.orangeUI
+        
         UISwitch.appearance().onTintColor = theme.colors.orangeUI
-        UISearchBar.appearance().barTintColor = .white
+        UISwitch.appearance().thumbTintColor = theme == PresentationTheme.darkTheme ? theme.colors.background : nil
+        UISwitch.appearance().tintColor = theme.colors.switchBackgroundColor
+        UISwitch.appearance().backgroundColor = theme.colors.switchBackgroundColor
+        UISwitch.appearance().cornerRadius = 16
```
carolanitz: I really don't think we need the cornerRadius

Sources/AppearanceManager.swift
```diff
@@ -50,3 +55,10 @@ extension UINavigationController {
         return self.topViewController
     }
 }
+
+extension UISwitch {
+    @objc dynamic var cornerRadius: CGFloat {
+        get { return layer.cornerRadius }
+        set { layer.cornerRadius = newValue }
+    }
+}
```
carolanitz: I personally find introducing this more confusing than helpful

Sources/VLCSettingsController.m
```diff
@@ -35,19 +46,33 @@ - (instancetype)initWithStyle:(UITableViewStyle)style
...
-    self.view.backgroundColor = PresentationTheme.current.colors.settingsBackground;
-    self.tableView.separatorColor = PresentationTheme.current.colors.settingsSeparatorColor;
+    self.view.backgroundColor = PresentationTheme.current.colors.background;
     [self.tableView reloadData];
     [self setNeedsStatusBarAppearanceUpdate];
+    
+    // Force UIKit to reload after changing UIAppearance
+    UIView *superView = self.tableView.superview;
+    [self.tableView removeFromSuperview];
+    [superView addSubview:self.tableView];
```
carolanitz: This looks wrong to me and like fixing a symptom instead of the underlying issue, which is that views don't reload after the appearance change. This would be a general problem that should be fixed at a central point or if we can't fix this we should think about not defining certain properties in UIAppearance

Sources/VLCSettingsController.m
```diff
@@ -125,12 +150,46 @@ - (void)updateUIAndCoreSpotlightForPasscodeSetting:(BOOL)passcodeOn
...
     [self.navigationController pushViewController:aboutVC animated:YES];
 }
 
+- (CGFloat) settingsViewController:(id<IASKViewController>)settingsViewController tableView:(UITableView *)tableView heightForHeaderForSection:(NSInteger)section {
+    return section == 0 ? 0 : SETTINGS_HEADER_HEIGHT;
+}
+
+- (UIView *)settingsViewController:(id<IASKViewController>)settingsViewController tableView:(UITableView *)tableView viewForHeaderForSection:(NSInteger)section
+{
+    UIView *header = [[UIView alloc] initWithFrame:CGRectMake(HEADER_OFFSET, 0, tableView.bounds.size.width, SETTINGS_HEADER_HEIGHT)];
```
carolanitz: Why are you not using a UITableViewHeaderFooterView? I know that this is only a tiny class but I'd still rather have that as it's own class

Sources/VLCSettingsController.m
```diff
@@ -35,19 +46,33 @@ - (instancetype)initWithStyle:(UITableViewStyle)style
...
     self.modalPresentationStyle = UIModalPresentationFormSheet;
     self.delegate = self;
     self.showDoneButton = NO;
     self.showCreditsFooter = NO;
+    self.neverShowPrivacySettings = YES;
+    
+    self.navigationItem.leftBarButtonItem = [[UIBarButtonItem alloc] initWithTitle:NSLocalizedString(@"BUTTON_ABOUT", nil) style:UIBarButtonItemStylePlain target:self action:@selector(pushAboutViewController:)];
+    self.navigationItem.leftBarButtonItem.accessibilityIdentifier = VLCAccessibilityIdentifier.about;
+    
+    self.tableView.estimatedRowHeight = MINIMUM_SETTINGS_ROW_HEIGHT;
```
carolanitz: I would set here also the rowheight to be consistent

Sources/VLCSettingsController.m
```diff
@@ -125,12 +150,46 @@ - (void)updateUIAndCoreSpotlightForPasscodeSetting:(BOOL)passcodeOn
...
+    UIView *separator = [[UIView alloc] initWithFrame:CGRectMake(HEADER_OFFSET, 0, header.frame.size.width - (HEADER_OFFSET * 2), 2)];
+    separator.backgroundColor = PresentationTheme.current.colors.settingsSeparatorColor;
+    [header addSubview:separator];
+    return header;
+}
+
+- (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    return UITableViewAutomaticDimension;
+}
```
carolanitz: can go away if you set the rowheight property up top

Sources/VLCSettingsController.m
```diff
@@ -125,12 +150,46 @@ - (void)updateUIAndCoreSpotlightForPasscodeSetting:(BOOL)passcodeOn
...
+    [NSLayoutConstraint activateConstraints:@[
+                                              [headerLabel.leadingAnchor constraintEqualToAnchor:margins.leadingAnchor constant:8],
+                                              [headerLabel.trailingAnchor constraintEqualToAnchor:margins.trailingAnchor],
+                                              [headerLabel.topAnchor constraintEqualToAnchor:margins.topAnchor],
+                                              [headerLabel.bottomAnchor constraintEqualToAnchor:margins.bottomAnchor]
+                                              ]];
+    
+    UIView *separator = [[UIView alloc] initWithFrame:CGRectMake(HEADER_OFFSET, 0, header.frame.size.width - (HEADER_OFFSET * 2), 2)];
+    separator.backgroundColor = PresentationTheme.current.colors.settingsSeparatorColor;
+    [header addSubview:separator];
```
carolanitz: This will "break" and by that I mean the separator will be too short when you rotate the view since you set a static frame

Sources/VLCSettingsController.m
```diff
@@ -171,4 +226,45 @@ - (void)updateForPasscode:(NSString *)passcode
...
     }
 }
+
+# pragma mark - IASKAppSettingsViewController override
+
+- (UITableViewCell*)newCellForIdentifier:(NSString*)identifier {
+    UITableViewCell *cell = nil;
+    if ([identifier isEqualToString:kIASKPSToggleSwitchSpecifier]) {
+        cell = [[UITableViewCell alloc] initWithStyle:UITableViewCellStyleSubtitle reuseIdentifier:kIASKPSToggleSwitchSpecifier];
+        cell.accessoryView = [[IASKSwitch alloc] initWithFrame:CGRectMake(0, 0, 79, 27)];
```
carolanitz: Can we write cell subclasses for this and not cram everything in this method and viewcontroller? Also magic numbers. By having subclass in an extra file we should also be able to remove some of the above imports I reckon, which will make the entire thing more readable

Sources/VLCSettingsController.m
```diff
@@ -171,4 +226,45 @@ - (void)updateForPasscode:(NSString *)passcode
         [self.navigationController.presentedViewController dismissViewControllerAnimated:YES completion:nil];
     }
 }
+
+# pragma mark - IASKAppSettingsViewController override
+
+- (UITableViewCell*)newCellForIdentifier:(NSString*)identifier {
```
carolanitz: space before the *

Sources/VLCSettingsController.m
```diff
@@ -196,4 +209,43 @@ - (void)updateForPasscode:(NSString *)passcode
...
+    mask.translatesAutoresizingMaskIntoConstraints = false;
+    [cell.contentView addSubview:mask];
+    
+    UILayoutGuide *margins = [cell.contentView layoutMarginsGuide];
+    [NSLayoutConstraint activateConstraints:@[
+                                              [mask.leadingAnchor constraintEqualToAnchor:margins.leadingAnchor],
+                                              [mask.trailingAnchor constraintEqualToAnchor:margins.trailingAnchor],
+                                              [mask.topAnchor constraintEqualToAnchor:margins.topAnchor],
+                                              [mask.bottomAnchor constraintEqualToAnchor:margins.bottomAnchor],
+                                              [mask.heightAnchor constraintGreaterThanOrEqualToConstant:MINIMUM_SETTINGS_ROW_HEIGHT]
```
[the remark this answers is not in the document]
carolanitz: no hacks, please. If you want to ensure the correct height that leaves enough room you overwrite a cells systemLayoutSizeFittingSize or something like that. There are a bunch of articles on how to achieve this. Your implementation will not leave enough room when you have large text which is what you wanted if I understand the automatic dimension correctly. This will also make the MinimumSettingsrowheight obsolete

Sources/VLCSettingsController.m
```diff
@@ -125,12 +138,38 @@ - (void)updateUIAndCoreSpotlightForPasscodeSetting:(BOOL)passcodeOn
 
 #pragma mark - IASKSettings delegate
 
-- (void)settingsViewController:(IASKAppSettingsViewController*)sender buttonTappedForSpecifier:(IASKSpecifier*)specifier
+- (IBAction)pushAboutViewController:(id)sender
```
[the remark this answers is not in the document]
mkchoi212: Don't know what I was thinking :p Rookie mistake!

Sources/AppearanceManager.swift
mkchoi212: If the cornerRadius is not set, the backgroundColor leaks outside the switch. The product looks something like this 
untitled

Sources/AppearanceManager.swift
mkchoi212: 🤔 would it be better to just set the cornerRadius locally wherever a switch is used?

Sources/VLCSettingsController.m
mkchoi212: It's been fixed with a `reloadSections`

Sources/VLCSettingsController.m
mkchoi212: duh 😄

Sources/AppearanceManager.swift
carolanitz: either we use it wrong or it's a UIKit bug and we should file a radar with a simple sample project that shows the issue. setting the corner radius could bite us one day if Apple changes the control from what it is right now

carolanitz: just did a quick search on Google and a lot run into that so leave it for now, file a radar and add a comment

Sources/AppearanceManager.swift
carolanitz: Something that you could do, which I would understand is this: 
``class VLCSwitch: UISwitch {
  var OffColor: UIColor? {
        didSet {
            self.layer.cornerRadius = 16 
            self.backgroundColor = OffColor
        }
    }
}

Sources/VLCSettingsController.m
mkchoi212: I tried what you suggested but I think this is necessary to override the `heightForRowAtIndexPath` already declared in its parent class `IASKSettingsViewController`

Sources/VLCSettingsTableHeaderView.swift
```diff
@@ -0,0 +1,65 @@
...
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import UIKit
+
+class VLCSettingsTableHeaderView: UITableViewHeaderFooterView {
+    @objc static let identifier = String(describing: VLCSettingsTableHeaderView.self)
+    @objc var headerLabel: UILabel!
+    
+    override var frame: CGRect {
```
mkchoi212: This was required to silence auto layout warnings. See https://stackoverflow.com/a/35053234/4064189

carolanitz: could you try giving the constraints a priority of 999. This allows them to break temporarily while the height or width is 0

Sources/VLCSettingsTableHeaderView.swift
```diff
@@ -0,0 +1,65 @@
...
+        separator.backgroundColor = PresentationTheme.current.colors.settingsSeparatorColor
+        separator.translatesAutoresizingMaskIntoConstraints = false
+        contentView.addSubview(separator)
+        
+        let margins = contentView.layoutMarginsGuide
+        let offset = textLabel!.frame.origin.x
+        NSLayoutConstraint.activate([
+            separator.leadingAnchor.constraint(equalTo: margins.leadingAnchor, constant: offset),
+            separator.trailingAnchor.constraint(equalTo: margins.trailingAnchor, constant: offset),
+            separator.heightAnchor.constraint(equalToConstant: 2),
```
carolanitz: separators usually have a height of 1 or 0.5 even why the 2?

Sources/VLCSettingsTableHeaderView.swift
```diff
@@ -0,0 +1,65 @@
...
+            separator.topAnchor.constraint(equalTo: margins.topAnchor)
+        ])
+        
+        headerLabel = UILabel(frame: contentView.frame)
+        headerLabel.translatesAutoresizingMaskIntoConstraints = false
+        headerLabel.font = UIFont.systemFont(ofSize: 23, weight: .semibold)
```
carolanitz: this font shouldn't be defined here

Sources/VLCSettingsTableViewCell.m
```diff
@@ -0,0 +1,73 @@
+/*****************************************************************************
```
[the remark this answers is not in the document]
carolanitz: also prepare for reuse is missing

Sources/VLCSettingsTableViewCell.m
```diff
@@ -0,0 +1,73 @@
...
+    self.mainLabel.textColor = PresentationTheme.current.colors.cellTextColor;
+    self.subtitleLabel.textColor = PresentationTheme.current.colors.cellDetailTextColor;
+    [super awakeFromNib];
+}
+
+- (void)configureForSpecifier:(IASKSpecifier *)specifier value:(id)currentValue
+{
+    self.mainLabel.text = specifier.title;
+    
+    if ([specifier.type isEqualToString:kIASKPSToggleSwitchSpecifier]) {
```
carolanitz: switch over the specifier.type

Sources/VLCSettingsController.m
```diff
@@ -146,16 +153,20 @@ - (void)PAPasscodeViewControllerDidSetPasscode:(PAPasscodeViewController *)contr
...
-    cell.backgroundColor = PresentationTheme.current.colors.settingsCellBackground;
-    cell.textLabel.textColor = PresentationTheme.current.colors.cellTextColor;
-    cell.detailTextLabel.textColor = PresentationTheme.current.colors.cellDetailTextColor;
-    if ([specifier.key isEqualToString:@"about"]) {
-        cell.accessibilityIdentifier = VLCAccessibilityIdentifier.about;
+    VLCSettingsTableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:specifier.type];
+    if (!cell) {
+        cell = [VLCSettingsTableViewCell cellWithIdentifier:specifier.type target:self];
     }
+    [cell configureForSpecifier:specifier value:[self.settingsStore objectForKey:specifier.key]];
```
carolanitz: value is too vague can we have a variable instead and a couple of checks that the object that you try to get out of there is really the type that you expect?

Sources/VLCSettingsTableViewCell.m
```diff
@@ -0,0 +1,73 @@
...
+@implementation VLCSettingsTableViewCell
+
++ (VLCSettingsTableViewCell *)cellWithIdentifier:(NSString *)identifier target:(id)target {
+    NSArray *nibArray = [[NSBundle mainBundle] loadNibNamed:@"VLCSettingsTableViewCell" owner:nil options:nil];
+    VLCSettingsTableViewCell *cell = (VLCSettingsTableViewCell *)[nibArray lastObject];
+
+    if ([identifier isEqualToString:kIASKPSToggleSwitchSpecifier]) {
+        cell.selectionStyle = UITableViewCellSelectionStyleNone;
+        cell.accessoryView = [[IASKSwitch alloc] initWithFrame:CGRectZero];
+        [(IASKSwitch *)(cell.accessoryView) addTarget:target action:NSSelectorFromString(@"toggledValue:") forControlEvents:UIControlEventValueChanged];
```
carolanitz: could we have the entire cell be a touch target and not the switch itself?

mkchoi212: Should the selection style still be .none or changed to .default?

Sources/VLCSettingsTableHeaderView.swift
mkchoi212: Figma mockups say 2 but changing it to 1 looks a lot better

Sources/VLCSettingsController.m
mkchoi212: All type checks are done by `configureForSpecifier` for different types of cells with functions from InAppSettingsKit :D

Sources/VLCSettingsController.m
```diff
@@ -42,14 +43,16 @@ - (void)viewDidLoad
...
     [self.tableView registerClass:[VLCSettingsTableHeaderView class] forHeaderFooterViewReuseIdentifier:VLCSettingsTableHeaderView.identifier];
     [self themeDidChange];
 }
 
 - (void)themeDidChange
 {
-    self.view.backgroundColor = PresentationTheme.current.colors.settingsBackground;
-    self.tableView.separatorColor = PresentationTheme.current.colors.settingsSeparatorColor;
+    self.view.backgroundColor = PresentationTheme.current.colors.background;
+    self.tableView.tintColor = PresentationTheme.current.colors.orangeUI;
```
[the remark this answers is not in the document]
mkchoi212: whoops!

Sources/AppearanceManager.swift
mkchoi212: I prefer the above approach as well but InAppSettingsKit is not letting me do this :(
They force cast `UISwitch` to their `IASKSwitch` in their library.

Sources/VLCSettingsController.m
```diff
@@ -172,4 +183,28 @@ - (void)updateForPasscode:(NSString *)passcode
         [self.navigationController.presentedViewController dismissViewControllerAnimated:YES completion:nil];
     }
 }
+
+#pragma mark - InAppSettings customization
+
+- (NSIndexPath *)tableView:(UITableView *)tableView willSelectRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    IASKSpecifier *specifier = [self.settingsReader specifierForIndexPath:indexPath];
+    return [specifier.type isEqual: kIASKPSSliderSpecifier] ? nil : indexPath;
```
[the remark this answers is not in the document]
mkchoi212: The `kIASKPSToggleSwitchSpecifier` is to also allow users to toggle the switch by pressing the cell. Is this not what you meant by `have the entire cell be a touch target and not the switch itself?` 🤔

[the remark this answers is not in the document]
carolanitz: oh yeah you're right if we include it here the event wouldn't go through. can we just make sure that it doesn't get the grey selectionstate when you select the cell?

carolanitz:
merged with ff9342f72f7e516a941a8c77afccb002106b4bfe and following

mkchoi212:
@all thanks for all the comments guys! this was a big one :D

PR #77, opened 2018-06-11
Refactor existing tests

mkchoi212:
### Checklist


.circleci/config.yml
```diff
@@ -19,16 +19,16 @@ jobs:
...
           command: |
-            curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf
-            pod install
+            [ -d "Pods" ] || (curl https://cocoapods-specs.circleci.com/fetch-cocoapods-repo-from-s3.sh | bash -s cf; pod install)
       - save_cache:
-          key: v1-pods-{{ checksum "Podfile.lock" }}
+          key: v2-pods-{{ checksum "Podfile" }}
           paths:
             - Pods
+            - Podfile.lock
```
carolanitz: do you happen to have a reference for this change by any chance ? I worked on that with the lead engineer from circleci and I wonder if there was a reason dantoml used the .lock file over the Podfile.

Sources/VLCMediaDataSource.swift
```diff
@@ -89,6 +89,7 @@ struct VLCMediaType {
             preconditionFailure("TODO")
             return videoPlaylist
         case .allVideos:
+            getAllVideos()
```
carolanitz: This fixes the test but at a very high expense. What you do here is that you call the generate all videos method for every single cell actually, because this method is called from object(at index: Int, subcategory: VLCMediaSubcategory). getAllVideos() is a very expensive method because it does fetchrequests to read all the videos out of the database with fetchrequests. 

How we would want to fix it, is by only regenerating the allVideos array when we detect that files were added or deleted. That's why there is a mediaFileDiscoverer and in the videostab we should have a method that is called with the changes that can then animate insertion or deletion of cells. That's at least how it should work ideally at the end. 
For a beginning it's enough if it refreshes when we detect videos have been added or deleted

VLC-iOS-UITests/Screenshot.swift
```diff
@@ -60,17 +60,17 @@ class Screenshot: XCTestCase {
         helper.tapTabBarItem(VLCAccessibilityIdentifier.localNetwork)
         app.cells[VLCAccessibilityIdentifier.downloads].tap()
 
-        let downloadTextfield = app.textFields["http://myserver.com/file.mkv"]
+        let downloadTextfield = app.textFields["http://myserver.com/file.mkv"].firstMatch
```
carolanitz: 👍

.circleci/config.yml
mkchoi212: I don't have any references but I noticed that after running `pod install`, the `PODFILE CHECKSUM` variable within `Podfile.lock` changes each time. Because of this, the cache-key used to save the cache is different than that used when loading the cache.

Oh but I just noticed this 😅
> If you are using CocoaPods, then we recommend that you check your Pods directory into source control. This will ensure that you have a deterministic, reproducible build.
https://circleci.com/docs/2.0/testing-ios/#using-cocoapods

Mikanbu:
Merged with 0c368fe0120dd75a4bcdc801d973e178705da16b!

PR #79, opened 2018-06-15
CI: Cache installed pods

mkchoi212:
### Checklist

### Description

According to CircleCI, 
> If you are using CocoaPods, then we recommend that you check your Pods directory into source control. This will ensure that you have a deterministic, reproducible build.
https://circleci.com/docs/2.0/testing-ios/#using-cocoapods

But `Pods` is 2.6 GB while the entire project is 2.8 GB 😅
So this PR tries to cache the pods in CircleCI instead of the repo with contents of `Podfile` as the key to the cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,16 +19,16 @@ jobs:
           paths:
             - .bundle
       - restore_cache:
-          key: v1-pods-{{ checksum "Podfile.lock" }}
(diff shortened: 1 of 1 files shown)


carolanitz:
this was merged with 9573d3a4cef1d696d20015e283adb4478e39e056

PR #83, opened 2018-06-22
VLCMovieViewController: Correctly display Playing Externally view

carolanitz:
### Checklist

### Description

 When the renderer is disconnected we want to hide the playing Externally view
 When we mirror the screen we don't want to show it either so we just check if we have a renderer

I haven't found a better way to detect if we're mirroring or not so I use if we have a renderer right now, which is not ideal. I'm also not sure if there are other cases where we display on an external screen but this fixes it for the two reported bugs


Mikanbu:
Merged by 8e664031eeac79003f7bc69df4e1310ec209e8ec.

PR #85, opened 2018-06-27
Fix Fill to screen(closes #209) 

Mikanbu:
### Checklist

### DIFF
--- a/Sources/VLCPlaybackController.m
+++ b/Sources/VLCPlaybackController.m
@@ -878,12 +878,6 @@ typedef NS_ENUM(NSUInteger, VLCAspectRatio) {
     return [[UIDevice currentDevice] VLCHasExternalDisplay] ? [UIScreen screens][1] : [UIScreen mainScreen];
 }
 
-- (NSString *)screenAspectRatio
-{
-    CGSize currentScreenSize = [self currentScreen].bounds.size;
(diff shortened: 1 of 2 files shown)


jbkempf:
LGTM

Sources/VLCPlaybackController.m
```diff
@@ -872,36 +873,65 @@ - (void)jumpBackward:(int)interval
...
+
+    if (dar >= ar) {
+        scale = screenSize.width / (float)videoSize.width;
+    } else {
+        scale = screenSize.height / (float)videoSize.height;
+    }
+
+    // Multiplied by screen.scale in consideration of pt to px
+    _mediaPlayer.scaleFactor = scale * screen.scale;
+    _isInFillToScreen = YES;
```
carolanitz: I don't think you need to introduce this state. Can't you just check the scaleFactor on the mediaPlayer?

Mikanbu: Well, I guess we could but it felt a bit wrong since the `scaleFactor` is not particularly reserved for the FillToScreen feature, it could be used for other things.
At the end of the day, I have a slight preference for the `BOOL` but I don't mind changing it.

Mikanbu:
Merged with 86162757 and 6c00fda6

PR #91, opened 2018-07-11
Settings: tableview section separator

mkchoi212:
### Checklist

### Description

Added a separator in-between tableview sections.

### DIFF
--- a/SharedSources/PresentationTheme.swift
+++ b/SharedSources/PresentationTheme.swift
@@ -76,6 +76,15 @@ extension Notification.Name {
     }
 }
 
+@objcMembers class Typography: NSObject {
+    
+    let tableHeaderFont: UIFont
+    
+    init(tableHeaderFont: UIFont) {
+        self.tableHeaderFont = tableHeaderFont
+    }
+}
+
 @objcMembers class PresentationTheme: NSObject {
 
     static let brightTheme = PresentationTheme(colors: brightPalette)
@@ -97,6 +106,7 @@ extension Notification.Name {
     }
 
     let colors: ColorPalette
+    let font = defaultFont
 }
 
 @objc extension UIColor {
@@ -171,3 +181,5 @@ let darkPalette = ColorPalette(isDark: true,
                                mediaCategorySeparatorColor: UIColor(0x25292C),
                                tabBarColor: UIColor(0x292B36),
                                orangeUI: UIColor(0xFF8800))
+
+let defaultFont = Typography(tableHeaderFont: UIFont.systemFont(ofSize: 24, weight: .semibold))
--- a/SharedSources/PresentationTheme.swift
+++ b/SharedSources/PresentationTheme.swift
@@ -30,7 +30,6 @@ extension Notification.Name {
     let cellTextColor: UIColor
     let lightTextColor: UIColor
     let sectionHeaderTextColor: UIColor
-    let sectionHeaderTintColor: UIColor
     let settingsBackground: UIColor
     let settingsCellBackground: UIColor
     let separatorColor: UIColor
@@ -50,7 +49,6 @@ extension Notification.Name {
                 cellTextColor: UIColor,
                 lightTextColor: UIColor,
                 sectionHeaderTextColor: UIColor,
-                sectionHeaderTintColor: UIColor,
                 settingsBackground: UIColor,
                 settingsCellBackground: UIColor,
                 separatorColor: UIColor,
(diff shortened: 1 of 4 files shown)


SharedSources/PresentationTheme.swift
```diff
@@ -79,6 +76,15 @@ extension Notification.Name {
     }
 }
 
+@objcMembers public class Typography: NSObject {
+    
+    public let tableHeader: UIFont
+    
+    public init(tableHeader: UIFont) {
+        self.tableHeader = tableHeader
```
Mikanbu: Is there a specific reason why we would want this class to be `public`?

Sources/VLCSettingsTableHeaderView.swift
```diff
@@ -0,0 +1,64 @@
...
+
+import UIKit
+
+class VLCSettingsTableHeaderView: UITableViewHeaderFooterView {
+    let headerHeight: CGFloat = 1.0
+    
+    @objc static let identifier = String(describing: VLCSettingsTableHeaderView.self)
+    @objc var headerLabel: UILabel!
+    @objc var separator: UIView!
+
```
Mikanbu: What do you think about initializing them directly here?
For example: 

```

    @objc var separator: UIView = {
        let separator = UIView()
        separator.backgroundColor = PresentationTheme.current.colors.separatorColor
        separator.translatesAutoresizingMaskIntoConstraints = false
        return separator
    }()
```
This will make the init clearer by only having the `addSubView()` calls and the constraints stuff.

Sources/VLCSettingsTableHeaderView.swift
```diff
@@ -0,0 +1,64 @@
...
+    let headerHeight: CGFloat = 1.0
+    
+    @objc static let identifier = String(describing: VLCSettingsTableHeaderView.self)
+    @objc var headerLabel: UILabel!
+    @objc var separator: UIView!
+    
+    override init(reuseIdentifier: String?) {
+        super.init(reuseIdentifier: reuseIdentifier)
+
+        constraints.forEach { $0.priority = UILayoutPriority(rawValue: 999) }
```
Mikanbu: I think it's better if we use `.required - 1` here.

Sources/VLCSettingsTableHeaderView.swift
mkchoi212: Good point. I always forget you can do this :D

SharedSources/PresentationTheme.swift
mkchoi212: Hmmm I guess it's unnecessary

Sources/VLCSettingsTableHeaderView.swift
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Authors: Mike JS. Choi <mkchoi212 # icloud.com>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import UIKit
+
+class VLCSettingsTableHeaderView: UITableViewHeaderFooterView {
+    let headerHeight: CGFloat = 1.0
```
Mikanbu: Do we need to expose this outside of this class? If not we should consider "privatising" it 😄

SharedSources/PresentationTheme.swift
```diff
@@ -79,6 +76,15 @@ extension Notification.Name {
     }
 }
 
+@objcMembers class Typography: NSObject {
+    
+    let tableHeader: UIFont
+    
+    init(tableHeader: UIFont) {
```
carolanitz: tableHeaderFont

Sources/VLCSettingsController.m
```diff
@@ -172,4 +174,27 @@ - (void)updateForPasscode:(NSString *)passcode
...
+    }
+    VLCSettingsTableHeaderView *header = [tableView dequeueReusableHeaderFooterViewWithIdentifier:VLCSettingsTableHeaderView.identifier];
+    header.headerLabel.text = [self.settingsReader titleForSection:section];
+    return header;
+}
+
+- (NSString *)tableView:(UITableView *)tableView titleForHeaderInSection:(NSInteger)section
+{
+    return nil;
+}
```
carolanitz: This can be deleted it will fall back to viewForHeaderInSection

Sources/VLCSettingsTableHeaderView.swift
```diff
@@ -0,0 +1,70 @@
...
+ * $Id$
+ *
+ * Authors: Mike JS. Choi <mkchoi212 # icloud.com>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import UIKit
+
+class VLCSettingsTableHeaderView: UITableViewHeaderFooterView {
```
carolanitz: Could you adjust this to the new class VLCSectionTableHeaderView?

Sources/VLCSettingsController.m
mkchoi212: Hmmm without it, the header titles come out capitalized. I think this is because InAppSettings has a default `titleForHeaderInSection`. So this is required to hide `super`'s implementation.

carolanitz:
merged with 7ea70270b0f31fbf6b72f6b33f5c0cabc3190c5c and following

PR #97, opened 2018-07-13
[WIP]: Add edit feature

Mikanbu:
### Checklist

### Description

This PR aims to add the edit feature.
Most of the value such as: sizes, spacing, etc... comes from [figma](https://www.figma.com/file/ByhPBHr4gTmUwo4bsVcanWIN/VLC180510?node-id=0%3A2 ).
Currently, it has the layout needed for the cells and **needs**:

- Add little check mark inside the orange circle
- Handle Swipe
- Hande Subcategory & tabbar
- Link to data
- Handle theme changes
- Add toolbar
- Handle video and audio thumbnail size(I didn't realize there was a difference (-ω-、))


Sources/MediaSubcategoryViewController.swift
```diff
@@ -67,8 +68,12 @@ class VLCMediaSubcategoryViewController: BaseButtonBarPagerTabStripViewControlle
         if #available(iOS 11.0, *) {
             navigationController?.navigationBar.prefersLargeTitles = false
         }
+
+        let fixedSpace: UIBarButtonItem = UIBarButtonItem(barButtonSystemItem: .fixedSpace, target: nil, action: nil)
+        fixedSpace.width = fixedSpaceWidth
+
         navigationItem.leftBarButtonItem = UIBarButtonItem(title: NSLocalizedString("SORT", comment: ""), style: .plain, target: self, action: #selector(sort))
-        navigationItem.rightBarButtonItem = UIBarButtonItem(customView: rendererButton)
+        navigationItem.rightBarButtonItems = [editButtonItem, fixedSpace, UIBarButtonItem(customView: rendererButton)]
```
carolanitz: Can we hide the rendererButton in editmode ?

Sources/MediaViewController.swift
```diff
@@ -164,25 +185,46 @@ class VLCMediaViewController<T>: UICollectionViewController, UICollectionViewDel
...
     override func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
-        if let playlistCell = collectionView.dequeueReusableCell(withReuseIdentifier: VLCPlaylistCollectionViewCell.cellIdentifier(), for: indexPath) as? VLCPlaylistCollectionViewCell {
-            if let mediaObject = subcategory.files[indexPath.row] as? NSManagedObject {
-                playlistCell.mediaObject = mediaObject
+        if isEditing {
+            if let cell = collectionView.dequeueReusableCell(withReuseIdentifier: VLCMediaViewEditCell.identifier,
+                                                             for: indexPath) as? VLCMediaViewEditCell {
+                cell.title.text = "( ｀ー´)ノ"
+                cell.subInfo.text = "(-ω-、)"
+                cell.size.text = "|ω°•)"
```
carolanitz: debugcode ?

Sources/MediaViewController.swift
```diff
@@ -164,25 +185,46 @@ class VLCMediaViewController<T>: UICollectionViewController, UICollectionViewDel
     }
 
     override func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
-        if let playlistCell = collectionView.dequeueReusableCell(withReuseIdentifier: VLCPlaylistCollectionViewCell.cellIdentifier(), for: indexPath) as? VLCPlaylistCollectionViewCell {
-            if let mediaObject = subcategory.files[indexPath.row] as? NSManagedObject {
-                playlistCell.mediaObject = mediaObject
+        if isEditing {
```
carolanitz: could we set a different datasource when we enter editmode to avoid the if else?

Sources/MediaViewController.swift
```diff
@@ -164,25 +185,46 @@ class VLCMediaViewController<T>: UICollectionViewController, UICollectionViewDel
...
             }
-            return playlistCell
         }
         return UICollectionViewCell()
     }
 
     // MARK: - UICollectionViewDelegate
     override func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
-        if let mediaObject = subcategory.files[indexPath.row] as? NSManagedObject {
+        if isEditing {
```
carolanitz: same here if we have a different datasource this code will become cleaner

Sources/MediaViewController.swift
```diff
@@ -164,25 +185,46 @@ class VLCMediaViewController<T>: UICollectionViewController, UICollectionViewDel
...
+            }
+        } else if let mediaObject = subcategory.files[indexPath.row] as? NSManagedObject {
             delegate?.mediaViewControllerDidSelectMediaObject(self, mediaObject: mediaObject)
         }
     }
 
     // MARK: - UICollectionViewDelegateFlowLayout
     func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, sizeForItemAt indexPath: IndexPath) -> CGSize {
 
+        if isEditing {
```
carolanitz: same with the delegate

Sources/VLCMediaViewEditCell.swift
```diff
@@ -0,0 +1,131 @@
...
+}
+
+class VLCMediaViewEditCell: UICollectionViewCell {
+
+    static let identifier = String(describing: VLCMediaViewEditCell.self)
+
+    static let height: CGFloat = 56
+
+    var checkView: VLCCheckView = {
+        var view = UIView(frame: CGRect(x: 0, y: 0, width: 20, height: 20))
```
carolanitz: CGRectZero should also work :)

Sources/VLCMediaViewEditCell.swift
```diff
@@ -0,0 +1,131 @@
...
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.clipsToBounds = true
+        view.layer.cornerRadius = view.frame.width / 2
+        view.layer.borderColor = UIColor.lightGray.cgColor
+        view.layer.borderWidth = 1
```
carolanitz: All of this belongs to me in the checkView

Sources/VLCMediaViewEditCell.swift
```diff
@@ -0,0 +1,131 @@
...
+
+    let subInfo: UILabel = {
+        let subInfo = UILabel()
+        subInfo.textColor = PresentationTheme.current.colors.cellTextColor
+        subInfo.font = UIFont.systemFont(ofSize: 13)
+        subInfo.translatesAutoresizingMaskIntoConstraints = false
+        return subInfo
+    }()
+
+    let size: UILabel = {
```
carolanitz: can we postfix all of those with label

Sources/VLCMediaViewEditCell.swift
```diff
@@ -0,0 +1,131 @@
...
+    }()
+
+    let mediaInfoStackView: UIStackView = {
+        let mediaInfoStackView = UIStackView()
+        mediaInfoStackView.spacing = 5.0
+        mediaInfoStackView.axis = .vertical
+        mediaInfoStackView.alignment = .leading
+        mediaInfoStackView.translatesAutoresizingMaskIntoConstraints = false
+        return mediaInfoStackView
+    }()
```
carolanitz: Do all of these really need to be properties? I can imagine that the stackviews could just be declared locally

Sources/VLCMediaViewEditCell.swift
```diff
@@ -0,0 +1,131 @@
...
+        addSubview(mainStackView)
+
+        var guide: LayoutAnchorContainer = self
+
+        if #available(iOS 11.0, *) {
+            guide = safeAreaLayoutGuide
+        }
+        NSLayoutConstraint.activate([
+            checkView.view.heightAnchor.constraint(equalToConstant: 20),
+            checkView.view.widthAnchor.constraint(equalTo: checkView.view.heightAnchor),
```
carolanitz: The checkview should resize itself according to the image

Sources/VLCMediaViewEditCell.swift
```diff
@@ -0,0 +1,131 @@
...
+
+            thumbnail.heightAnchor.constraint(equalToConstant: VLCMediaViewEditCell.height),
+            thumbnail.widthAnchor.constraint(equalTo: thumbnail.heightAnchor),
```
carolanitz: the ratio is different for video vs. audio files

Sources/VLCMediaViewEditCell.swift
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+    var view: UIView
+}
+
+class VLCMediaViewEditCell: UICollectionViewCell {
+
+    static let identifier = String(describing: VLCMediaViewEditCell.self)
+
+    static let height: CGFloat = 56
```
carolanitz: This should ideally not be necessary and the thumbnail size + contraints to the border should resize the cell as needed

Sources/VLCMediaViewEditCell.swift
Mikanbu: Not according to figma, is there a specific reason why?

Sources/MediaViewController.swift
Mikanbu: Indeed it sounds nicer! Should it be a custom dataSource? 🤔

Sources/VLCMediaViewEditCell.swift
carolanitz: this has less to do with figma but more with how the VLCCheckMark view should behave.

Sources/VLCMediaViewEditCell.swift
carolanitz: sizelabel, subinfolabel etc. :)

Sources/VLCMediaViewEditCell.swift
Mikanbu: Why should it resize? The only reason I possibly see is accessibility.

Sources/MediaViewController.swift
carolanitz: yes a custom object. you can take a look at the mediaviewcontroller where we switch between searchdatasource and the usual one

Sources/MediaViewController.swift
carolanitz: okay instead I would have a media object that get's injected looking at that

Sources/VLCMediaViewEditCell.swift
Mikanbu: Ah okay! Yes that can be done!

Sources/VLCMediaViewEditCell.swift
carolanitz: Easier to maintain in the future if we change the image for example or Apple does something crazy.

Mikanbu:
The base of the edit UI will be addressed by the medialibrary branch.
Another PR for the UI will be needed.

PR #100, opened 2018-07-19
Vlc open in fix

carolanitz:
### Checklist

### Description

This fixes all the open in Issues while still working with the Files app

### DIFF
--- a/Sources/VLCAppDelegate.m
+++ b/Sources/VLCAppDelegate.m
@@ -326,7 +326,7 @@ didFailToContinueUserActivityWithType:(NSString *)userActivityType
             NSURLRelationship relationship;
             [manager getRelationship:&relationship ofDirectoryAtURL:[NSURL fileURLWithPath:directoryPath] toItemAtURL:url error:&theError];
             if (relationship == NSURLRelationshipContains) {
-                [self playWithURL:url successCallback:nil errorCallback:nil];
+                [self playWithURL:url completion:nil];
             }
         } else if ([url.scheme isEqualToString:@"vlc-x-callback"] || [url.host isEqualToString:@"x-callback-url"]) {
             // URL confirmes to the x-callback-url specification
@@ -353,7 +353,18 @@ didFailToContinueUserActivityWithType:(NSString *)userActivityType
                     errorCallback = [NSURL URLWithString:value];
             }
             if ([action isEqualToString:@"stream"] && movieURL) {
-                [self playWithURL:movieURL successCallback:successCallback errorCallback:errorCallback];
+                [self playWithURL:movieURL completion:^(BOOL success) {
+                    NSURL *callback = success ? successCallback : errorCallback;
+                    if (@available(iOS 10, *)) {
+                         [[UIApplication sharedApplication] openURL:callback options:@{} completionHandler:nil];
+                    } else {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+                        /* UIApplication's replacement calls require iOS 10 or later, which we can't enforce as of yet */
+                       [[UIApplication sharedApplication] openURL:callback];
+#pragma clang diagnostic pop
+                    }
+                }];
             }
             else if ([action isEqualToString:@"download"] && movieURL) {
                 [self downloadMovieFromURL:movieURL fileNameOfMedia:fileName];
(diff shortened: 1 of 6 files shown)


Sources/VLCPlaybackController.m
```diff
@@ -142,8 +143,14 @@ - (void)openVideoSubTitlesFromFile:(NSString *)pathToFile
     [_mediaPlayer addPlaybackSlave:[NSURL fileURLWithPath:pathToFile] type:VLCMediaPlaybackSlaveTypeSubtitle enforce:YES];
 }
 
-- (void)playMediaList:(VLCMediaList *)mediaList firstIndex:(NSInteger)index subtitlesFilePath:(NSString *)subsFilePath
+- (void)playMediaList:(VLCMediaList *)mediaList firstIndex:(NSInteger)index subtitlesFilePath:( NSString * _Nullable )subsFilePath
```
Mikanbu: Cosmetic: extra spaces in subtitlesFilePath type~

Sources/VLCPlaybackController.m
```diff
@@ -142,8 +143,14 @@ - (void)openVideoSubTitlesFromFile:(NSString *)pathToFile
...
 {
+    [self playMediaList: mediaList firstIndex: index subtitlesFilePath: subsFilePath completion: nil];
+}
+
+- (void)playMediaList:(VLCMediaList *)mediaList firstIndex:(NSInteger)index subtitlesFilePath:( NSString * _Nullable )subsFilePath completion:(void (^ __nullable)(BOOL success))completion
```
Mikanbu: Same~

Sources/VLCDocumentClass.swift
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Authors: Carola Nitz <nitz.carola # googlemail.com>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import Foundation
+
+@objc(VLCDocumentClass)
+class DocumentClass: UIDocument {
```
Mikanbu: Do we need this class or was it to prepare for future checks?

Sources/VLCAppDelegate.m
```diff
@@ -353,7 +343,25 @@ - (BOOL)application:(UIApplication *)app openURL:(NSURL *)url options:(NSDiction
...
+                        }
+                    } else {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+                        /* UIApplication's replacement calls require iOS 10 or later, which we can't enforce as of yet */
+                        if (success) {
+                            [[UIApplication sharedApplication] openURL:successCallback];
+                        } else {
+                            [[UIApplication sharedApplication] openURL:errorCallback];
+                        }
```
Mikanbu: How would you feel about simplifying the success checks by caching the callback?
Something like:

```
NSURL *callback = success ? successCallback : errorCallback;
if (@available(iOS 10, *)) {
    [[UIApplication sharedApplication] openURL:callback options:@{} completionHandler:nil];
} else {
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
    /* UIApplication's replacement calls require iOS 10 or later, which we can't enforce as of yet */
    [[UIApplication sharedApplication] openURL:callback];
#pragma clang diagnostic pop
```

Sources/VLCPlaybackController.h
```diff
@@ -13,7 +13,8 @@
 
 #import "VLCEqualizerView.h"
 
-extern NSString *const VLCPlaybackControllerPlaybackDidStart;
+NS_ASSUME_NONNULL_BEGIN
```
Mikanbu: Shouldn't the nonnull be another commit which would come before this one? I feel that it's kind of wrong that this commit touches `VLCRendererDiscovererManager`.

Sources/VLCDocumentClass.swift
carolanitz: we need to have a subclass as mentioned in the Documentation. You can't use it directly

Sources/VLCAppDelegate.m
carolanitz: yes definitely!

Sources/VLCDocumentClass.swift
Mikanbu: I see, thanks for the explanation!

PR #102, opened 2018-07-23
Settings: Actionsheet

mkchoi212:
### Checklist

### Description

Must be rebased after the settings PR is merged

Instead of pushing a new view to adjust the settings, VLCActionSheet is used.

### DIFF
--- a/Sources/VLCActionSheet/VLCActionSheetCell.swift
+++ b/Sources/VLCActionSheet/VLCActionSheetCell.swift
@@ -11,7 +11,9 @@
 
 class VLCActionSheetCell: UICollectionViewCell {
 
-    static let identifier = "VLCActionSheetCell"
+    @objc static var identifier: String {
+        return String(describing: self)
+    }
 
     let icon: UIImageView = {
         let icon = UIImageView()
@@ -25,6 +27,7 @@ class VLCActionSheetCell: UICollectionViewCell {
         name.textColor = PresentationTheme.current.colors.cellTextColor
         name.font = UIFont.systemFont(ofSize: 15)
         name.translatesAutoresizingMaskIntoConstraints = false
+        name.setContentHuggingPriority(.defaultLow, for: .horizontal)
         return name
     }()
 
@@ -62,9 +65,6 @@ class VLCActionSheetCell: UICollectionViewCell {
         NSLayoutConstraint.activate([
             icon.heightAnchor.constraint(equalToConstant: 25),
             icon.widthAnchor.constraint(equalTo: icon.heightAnchor),
-
-            name.trailingAnchor.constraint(equalTo: stackView.trailingAnchor),
-            name.centerYAnchor.constraint(equalTo: stackView.centerYAnchor),
 
             stackView.leadingAnchor.constraint(equalTo: guide.leadingAnchor, constant: 20),
             stackView.trailingAnchor.constraint(equalTo: guide.trailingAnchor, constant: -20),
--- a/Sources/VLCActionSheet/VLCActionSheetSectionHeader.swift
+++ b/Sources/VLCActionSheet/VLCActionSheetSectionHeader.swift
@@ -15,7 +15,7 @@ class VLCActionSheetSectionHeader: UIView {
 
     let title: UILabel = {
         let title = UILabel()
-        title.font = UIFont.systemFont(ofSize: 13)
+        title.font = UIFont.boldSystemFont(ofSize: 13)
         title.translatesAutoresizingMaskIntoConstraints = false
         return title
     }()
--- a/Sources/VLCSettingsController.m
+++ b/Sources/VLCSettingsController.m
@@ -22,6 +22,10 @@ CGFloat const SETTINGS_HEADER_HEIGHT = 64.;
 NSString * const kVLCSectionTableHeaderViewIdentifier = @"VLCSectionTableHeaderViewIdentifier";
 
 @interface VLCSettingsController ()<PAPasscodeViewControllerDelegate>
+{
+    VLCActionSheet *actionSheet;
+    VLCSettingsSpecifierManager *specifierManager;
+}
 @end
 
 @implementation VLCSettingsController
@@ -64,11 +68,16 @@ NSString * const kVLCSectionTableHeaderViewIdentifier = @"VLCSectionTableHeaderV
     self.tableView.rowHeight = UITableViewAutomaticDimension;
     [self.tableView registerClass:[VLCSectionTableHeaderView class] forHeaderFooterViewReuseIdentifier:kVLCSectionTableHeaderViewIdentifier];
     [self themeDidChange];
+    
+    actionSheet = [[VLCActionSheet alloc] init];
+    actionSheet.modalPresentationStyle = UIModalPresentationCustom;
+    [actionSheet.collectionView registerClass:[VLCSettingsSheetCell class] forCellWithReuseIdentifier:VLCSettingsSheetCell.identifier];
+    
+    specifierManager = [[VLCSettingsSpecifierManager alloc] initWithSettingsReader:self.settingsReader settingsStore:self.settingsStore];
 }
 
 - (void)themeDidChange
 {
-
     self.view.backgroundColor = PresentationTheme.current.colors.background;
     [self setNeedsStatusBarAppearanceUpdate];
 }
@@ -209,9 +218,32 @@ NSString * const kVLCSectionTableHeaderViewIdentifier = @"VLCSectionTableHeaderV
     return header;
 }
 
+- (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    [tableView deselectRowAtIndexPath:indexPath animated:YES];
+    IASKSpecifier *specifier = [self.settingsReader specifierForIndexPath:indexPath];
+
+    if ([specifier.type isEqualToString: kIASKPSMultiValueSpecifier]) {
+        [self displayActionSheetFor:specifier];
+    } else {
+        [super tableView:tableView didSelectRowAtIndexPath:indexPath];
+    }
+}
+
 - (NSString *)tableView:(UITableView *)tableView titleForHeaderInSection:(NSInteger)section
 {
     return nil;
+}
+
+- (void)displayActionSheetFor:(IASKSpecifier *)specifier
+{
+    specifierManager.specifier = specifier;
+    actionSheet.delegate = specifierManager;
+    actionSheet.dataSource = specifierManager;
+    
+    [self presentViewController:actionSheet animated:YES completion:^{
+        [self->actionSheet.collectionView selectItemAtIndexPath:self->specifierManager.selectedIndex animated:NO scrollPosition:UICollectionViewScrollPositionCenteredVertically];
+    }];
 }
 
 @end
--- a/Sources/VLCSettingsSheetCell.swift
+++ b/Sources/VLCSettingsSheetCell.swift
@@ -1,0 +1,47 @@
+/*****************************************************************************
+ * VLCSettingSheetCell.swift
+ * VLC for iOS
+ *****************************************************************************
+ * Copyright © 2018 VLC authors and VideoLAN
+ * $Id$
+ *
+ * Authors: Mike JS. Choi <mkchoi212 # icloud.com>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+class VLCSettingsSheetCell: VLCActionSheetCell {
+
+    override var isSelected: Bool {
+        didSet {
+            let colors = PresentationTheme.current.colors
+            name.textColor = isSelected ? colors.orangeUI : colors.cellTextColor
+            tintColor = isSelected ? colors.orangeUI : colors.cellDetailTextColor
+            checkmark.isHidden = !isSelected
+        }
+    }
+    
+    let checkmark: UILabel = {
+        let checkmark = UILabel()
+        checkmark.text = "✓"
+        checkmark.font = UIFont.systemFont(ofSize: 18)
+        checkmark.textColor = PresentationTheme.current.colors.orangeUI
+        checkmark.translatesAutoresizingMaskIntoConstraints = false
+        checkmark.isHidden = true
+        return checkmark
+    }()
+    
+    override init(frame: CGRect) {
+        super.init(frame: frame)
+        setupViews()
+    }
+    
+    required init?(coder aDecoder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    private func setupViews() {
+        stackView.removeArrangedSubview(icon)
+        stackView.addArrangedSubview(checkmark)
+    }
+}
--- a/Sources/VLCSettingsSpecifierManager.swift
+++ b/Sources/VLCSettingsSpecifierManager.swift
@@ -1,0 +1,84 @@
+/*****************************************************************************
+ * VLCSettingsSpecifierManager.swift
+ * VLC for iOS
+ *****************************************************************************
+ * Copyright © 2018 VLC authors and VideoLAN
+ * $Id$
+ *
+ * Authors: Mike JS. Choi <mkchoi212 # icloud.com>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import UIKit
+
+class VLCSettingsSpecifierManager: NSObject {
+    
+    @objc var specifier: IASKSpecifier?
+    var settingsReader: IASKSettingsReader
+    var settingsStore: IASKSettingsStore
+    
+    var items: NSArray {
+        guard let items = specifier?.multipleValues() as NSArray? else {
+            fatalError("VLCSettingsSpecifierManager: No rows provided for \(specifier?.key() ?? "null specifier")")
+        }
+        return items
+    }
+    
+    @objc var selectedIndex: IndexPath {
+        let index: Int
+        if let selectedItem = settingsStore.object(forKey: specifier?.key()) {
+            index = items.index(of: selectedItem)
+        } else if let specifier = specifier {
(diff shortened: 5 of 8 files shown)


Sources/VLCActionSheet/VLCActionSheetCell.swift
```diff
@@ -19,12 +19,23 @@ class VLCActionSheetCell: UICollectionViewCell {
         icon.contentMode = .scaleAspectFit
         return icon
     }()
+    
+    let checkmark: UILabel = {
+        let checkmark = UILabel()
+        checkmark.text = "✓"
+        checkmark.font = UIFont.systemFont(ofSize: 15)
```
mkchoi212: Should this be bigger?

Mikanbu: Indeed, It's a bit small but I wonder if this shouldn't be a separate class. For example perhaps a `VLCSettingSheetCell`?

Mikanbu: Also, open thought, I'm aware that on Figma the checkmark is on the right but shouldn't we consider putting the check mark on the left like the edit mode? Might just be a personal preference but I think it will be nicer 😅

Mikanbu: Not all ActionSheet cells need/will need checkmarks.

mkchoi212: So for the few that need both checkmarks and images, it should be placed on the right, no?

mkchoi212: Just subclassed it to `VLCSettingSheetCell` and it makes more sense :D

Sources/VLCActionSheet/VLCActionSheetCell.swift
```diff
@@ -11,7 +11,9 @@
 
 class VLCActionSheetCell: UICollectionViewCell {
 
-    static let identifier = "VLCActionSheetCell"
+    @objc static var identifier: String {
+        return String(describing: self)
```
carolanitz: That's interesting! Today I learned :) We should do that everywhere

Sources/VLCSettingsController.m
```diff
@@ -199,9 +201,43 @@ - (UIView *)settingsViewController:(id<IASKViewController>)settingsViewControlle
...
 
+- (void)displayActionSheetFor:(IASKSpecifier *)specifier
+{
+    VLCActionSheet *actionSheet = [[VLCActionSheet alloc] init];
+    actionSheet.modalPresentationStyle = UIModalPresentationCustom;
+    [actionSheet.collectionView registerClass:[VLCSettingsSheetCell class] forCellWithReuseIdentifier:VLCSettingsSheetCell.identifier];
+    
+    specifierManager = [[VLCSettingsSpecifierManager alloc] initWithSpecifier:specifier settingsReader:self.settingsReader settingsStore:self.settingsStore];
+    actionSheet.delegate = specifierManager;
+    actionSheet.dataSource = specifierManager;
```
carolanitz: It should be enough to create the Actionsheet and specifierManager once. Right now you do it on every tab of the cell.

Sources/VLCSettingsController.m
```diff
@@ -199,9 +201,43 @@ - (UIView *)settingsViewController:(id<IASKViewController>)settingsViewControlle
...
+    
+    [actionSheet setActionWithClosure:^(id item) {
```
carolanitz: can we use something else than id here?

Sources/VLCSettingsController.m
```diff
@@ -199,9 +201,43 @@ - (UIView *)settingsViewController:(id<IASKViewController>)settingsViewControlle
...
+        [self.settingsStore setObject:item forKey:specifier.key];
+        [self.settingsStore synchronize];
+        NSDictionary *info = [NSDictionary dictionaryWithObject:item forKey:[specifier key]];
+        [[NSNotificationCenter defaultCenter] postNotificationName:kIASKAppSettingChanged
+                                                            object:[specifier key]
+                                                          userInfo:info];
```
carolanitz: do you need to fire the notification here? Is it not done in the setter of the settingsstore?

Sources/VLCSettingsTableViewCell.swift
```diff
@@ -30,12 +30,10 @@ class VLCSettingsTableViewCell: UITableViewCell {
...
             toggle.addTarget(target, action: #selector(target.toggledValue(_:)), for: .valueChanged)
             accessoryView = toggle
-            selectionStyle = .none
-        case kIASKOpenURLSpecifier, kIASKPSMultiValueSpecifier:
+        case kIASKOpenURLSpecifier:
             accessoryType = .disclosureIndicator
-            selectionStyle = .default
         default:
-            assertionFailure("\(reuseIdentifier) has not been defined for VLCSettingsTableViewCell")
+            accessoryType = .none
```
carolanitz: Can we handle it explicitly for the known cases and keep the assertionFailure?

Sources/VLCSettingsSpecifierManager.swift
```diff
@@ -0,0 +1,107 @@
...
+    
+    var items: NSArray {
+        guard let items = specifier.multipleValues() as NSArray? else {
+            fatalError("VLCSettingsSpecifierManager: No rows provided for \(specifier.key())")
+        }
+        return items
+    }
+    
+    var selectedIndex: IndexPath {
+        var index: Int
```
carolanitz: you can make that a let since it will definitely be set in the if else part

Sources/VLCSettingsSpecifierManager.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+    
+    @objc init(specifier: IASKSpecifier, settingsReader: IASKSettingsReader, settingsStore: IASKSettingsStore) {
+        self.specifier = specifier
+        self.settingsReader = settingsReader
+        self.settingsStore = settingsStore
+    }
+    
+    fileprivate func updateCollectionViewCellApparence(cell: VLCSettingsSheetCell, highlighted: Bool) {
+        var textColor = PresentationTheme.current.colors.cellTextColor
```
carolanitz: what do you think about this:
```
let colors = PresentationTheme.current.colors

cell.name.textColor = highlighted ? colors.orangeUI : colors.cellTextColor
cell.tintColor = highlighted ? colors.orangeUI : colors.cellDetailTextColor
        
cell.checkmark.isHidden = !highlighted
```

Sources/VLCSettingsSpecifierManager.swift
```diff
@@ -0,0 +1,107 @@
...
+            return
+        }
+
+        guard let oldCell = collectionView.cellForItem(at: selectedIndex) as? VLCSettingsSheetCell,
+            let newCell = collectionView.cellForItem(at: indexPath) as? VLCSettingsSheetCell else {
+                assertionFailure("VLCSettingsSpecifierManager: VLCActionSheetDelegate: Cell is not a VLCSettingsSheetCell")
+                return
+        }
+
+        updateCollectionViewCellApparence(cell: oldCell, highlighted: false)
```
carolanitz: instead of highlighted does the cell maybe know that it's highlighted or selected itself? Maybe you can overwrite the setter if that's the case

Sources/VLCSettingsSpecifierManager.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+    
+    func actionSheet(collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        guard let cell = collectionView.dequeueReusableCell(withReuseIdentifier: VLCSettingsSheetCell.identifier, for: indexPath) as? VLCSettingsSheetCell else {
+            assertionFailure("VLCSettingsSpecifierManager: VLCActionSheetDataSource: Unable to dequeue reusable cell")
+            return UICollectionViewCell()
+        }
+        
+        let titles = specifier.multipleTitles()
+        cell.name.text = settingsReader.title(forStringId: titles?[indexPath.row] as? String)
```
carolanitz: can we add a check for indexPath.row < titles.count? Otherwise we crash here and maybe as well that the object is of type String otherwise we fail here silently

Sources/VLCSettingsSheetCell.swift
```diff
@@ -0,0 +1,29 @@
+/*****************************************************************************
+ * VLCSettingSheetCell.swift
+ *
+ * Copyright © 2018 VLC authors and VideoLAN
+ * Copyright © 2018 Videolabs
```
carolanitz: I think you can remove the Videolabs copyright ;)

Sources/VLCSettingsController.m
```diff
@@ -199,9 +201,43 @@ - (UIView *)settingsViewController:(id<IASKViewController>)settingsViewControlle
...
+    actionSheet.modalPresentationStyle = UIModalPresentationCustom;
+    [actionSheet.collectionView registerClass:[VLCSettingsSheetCell class] forCellWithReuseIdentifier:VLCSettingsSheetCell.identifier];
+    
+    specifierManager = [[VLCSettingsSpecifierManager alloc] initWithSpecifier:specifier settingsReader:self.settingsReader settingsStore:self.settingsStore];
+    actionSheet.delegate = specifierManager;
+    actionSheet.dataSource = specifierManager;
+    
+    [actionSheet setActionWithClosure:^(id item) {
+        [self.settingsStore setObject:item forKey:specifier.key];
+        [self.settingsStore synchronize];
```
carolanitz: Doesn't the self here lead to a retain cycle?

Sources/VLCSettingsSpecifierManager.swift
```diff
@@ -0,0 +1,107 @@
...
+        } else {
+            index = items.index(of: specifier.defaultValue())
+        }
+        return IndexPath(row: index, section: 0)
+    }
+    
+    @objc init(specifier: IASKSpecifier, settingsReader: IASKSettingsReader, settingsStore: IASKSettingsStore) {
+        self.specifier = specifier
+        self.settingsReader = settingsReader
+        self.settingsStore = settingsStore
```
carolanitz: shouldn't there be a super.init() ?

Sources/VLCSettingsController.m
mkchoi212: I don't think so because the closure is receiving bunch of different types as `item`. But even if we were able to type check `item` as something like `long`, InAppSettings doesn't provide a `set___` method for all types

Sources/VLCSettingsController.m
mkchoi212: Wow, I didn't know that :D

Sources/VLCSettingsSpecifierManager.swift
mkchoi212: I like it a lot ❤️

Sources/VLCSettingsTableViewCell.swift
mkchoi212: Yes! Don't know what I was thinking :p

Sources/VLCSettingsSpecifierManager.swift
mkchoi212: Good catch :D

Sources/VLCActionSheet/VLCActionSheetSectionHeader.swift
```diff
@@ -15,7 +15,7 @@ class VLCActionSheetSectionHeader: UIView {
 
     let title: UILabel = {
         let title = UILabel()
-        title.font = UIFont.systemFont(ofSize: 13)
+        title.font = UIFont.boldSystemFont(ofSize: 13)
```
Mikanbu: Is this change according to Figma?

Sources/VLCActionSheet/VLCActionSheetCell.swift
```diff
@@ -25,6 +27,7 @@ class VLCActionSheetCell: UICollectionViewCell {
         name.textColor = PresentationTheme.current.colors.cellTextColor
         name.font = UIFont.systemFont(ofSize: 15)
         name.translatesAutoresizingMaskIntoConstraints = false
+        name.setContentHuggingPriority(.defaultLow, for: .horizontal)
```
Mikanbu: Do we need to set `ContentHuggingPriority` ?

Sources/VLCActionSheet/VLCActionSheetCell.swift
```diff
@@ -47,7 +50,7 @@ class VLCActionSheetCell: UICollectionViewCell {
         setupViews()
     }
 
-    private func setupViews() {
+    func setupViews() {
```
Mikanbu: I don't think this method should be internal.

Sources/VLCSettingsSpecifierManager.swift
```diff
@@ -0,0 +1,84 @@
...
+        } else {
+            fatalError("VLCSettingsSpecifierManager: No specifier provided")
+        }
+        return IndexPath(row: index, section: 0)
+    }
+    
+    @objc init(settingsReader: IASKSettingsReader, settingsStore: IASKSettingsStore) {
+        super.init()
+        self.settingsReader = settingsReader
+        self.settingsStore = settingsStore
```
Mikanbu: Shouldn't the properties be initialized before calling `super.init()`? Therefore not needing to implicitly unwrap the properties on top?

Sources/VLCActionSheet/VLCActionSheetSectionHeader.swift
mkchoi212: yep!

Sources/VLCActionSheet/VLCActionSheetCell.swift
mkchoi212: This is to extend the name label all the way to the right so the checkmark can be at the rightmost edge. It can be moved to the subclass if it's breaking things.

Sources/VLCSettingsSpecifierManager.swift
mkchoi212: ah good catch. completely missed this :D

Sources/VLCActionSheet/VLCActionSheetCell.swift
mkchoi212: can I ask you why??

Mikanbu: I believe that it would be quite wrong to let other classes have the possibility to initialize the properties of `VLCActionSheetCell`.

I'm not saying that `VLCActionSheetCell` is perfect but I think classes should try to stay "self-contained".

For example, I think this would be nicer:

```
class TestCell: VLCActionSheetCell {
        super.init(frame: frame) // super.setupViews() is called in the super.init()
        setupViews()
    }

    required init?(coder aDecoder: NSCoder) {
        fatalError("init(coder:) has not been implemented")
    }

    private func setupViews() {
        stackView.removeArrangedSubview(icon)
        stackView.addArrangedSubview(//something\\)
    }
}
```

Also, I'm sorry if I don't make any sense 😅

mkchoi212: Gotcha! And it totally makes sense :D

mkchoi212:
Just tested it with Chromecast and nothing seems to be broken @carolanitz 
> Btw, I deleted the "if only 1 renderer, automatically select just to get this screenshot

simulator screen shot - iphone x - 2018-08-01 at 10 57 43

mkchoi212:
🔔 ping!

Sources/VLCSettingsSpecifierManager.swift
```diff
@@ -0,0 +1,84 @@
...
+
+extension VLCSettingsSpecifierManager: VLCActionSheetDataSource {
+    
+    func numberOfRows() -> Int {
+        return items.count
+    }
+    
+    func actionSheet(collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        guard let cell = collectionView.dequeueReusableCell(withReuseIdentifier: VLCSettingsSheetCell.identifier, for: indexPath) as? VLCSettingsSheetCell else {
+            fatalError("VLCSettingsSpecifierManager: VLCActionSheetDataSource: Unable to dequeue reusable cell")
```
carolanitz: doesn't dequeueReusableCell return nil when there is no cell to reuse ? and shouldn't in that case a new Cell be created instead of having a fatalError ?

carolanitz:
@mkchoi212 looks fine to me just that one question regarding dequeueReusableCell :)

Sources/VLCSettingsSpecifierManager.swift
mkchoi212: Ah yes :D

mkchoi212:
🔔 ping :D

carolanitz:
sorry this time I saw something else instead of having fatalerrors which litterally crash the app we should probably return rather empty arrays and just have assertionfailures. The only time you should use fatalerror is when you corrupt data or get into an unrecoverable appstate where only a restart would be the solution. Other options are errormessages to the user

Sources/VLCSettingsSpecifierManager.swift
```diff
@@ -0,0 +1,84 @@
...
+
+extension VLCSettingsSpecifierManager: VLCActionSheetDataSource {
+    
+    func numberOfRows() -> Int {
+        return items.count
+    }
+    
+    func actionSheet(collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        guard let cell = collectionView.dequeueReusableCell(withReuseIdentifier: VLCSettingsSheetCell.identifier, for: indexPath) as? VLCSettingsSheetCell else {
+            return UICollectionViewCell()
```
carolanitz: might make sense to still have an assertionfailure here because this should only happen when someone forgot to register the cellidentifier and the cellclass right ?

carolanitz:
merged with 0bcf87a3b0662e8e6f16e6e4452f067ea87f37b8 and following

PR #104, opened 2018-07-23
VLCMovieViewController: fix displaying on external screen via HDMI

carolanitz:
### Checklist

### Description

 When we refactored we didn't pass the movieView to the external screen when displaying externally
This worked well for chromecast and mirroring but sadly not for attaching an external screen via HDMI

Still need to test this with mirroring and chromecast


carolanitz:
tested with screen mirroring :) looks good and behaves like Netflix with the movie only on TV and the controls on the phone

PR #109, opened 2018-07-31
Fix Settings view (closes #268)

mkchoi212:
### Checklist

### Description

Issue reference https://code.videolan.org/videolan/vlc-ios/issues/268.

From looking at https://developer.apple.com/documentation/uikit/uitableviewheaderfooterview/1624912-textlabel, it seems as if using the default `textLabel` to customize the header view may be a bad idea when also adding a separator (subview).

> Accessing the value in this property causes the view to create a default label for displaying a detail text string. If you are managing the content of the view yourself by adding subviews to the contentView property, you should not access this property.

### DIFF
--- a/Sources/VLCSectionTableHeaderView.swift
+++ b/Sources/VLCSectionTableHeaderView.swift
@@ -15,6 +15,21 @@ import Foundation
 class VLCSectionTableHeaderView: UITableViewHeaderFooterView {
 
     let separator = UIView()
+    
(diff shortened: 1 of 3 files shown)


carolanitz:
👍

carolanitz:
merged with b4f76c4063bd2e90ef296ca6f3cd1b9747642836 & 05c32527594d93641528960a0b76c3b481dc9cb2

PR #110, opened 2018-08-02
VLCURLHandler: Have a designated Class for handling openURL calls

carolanitz:
Add first UnitTests for the parts of the URLHandler as well

### Checklist


Podfile
```diff
@@ -29,6 +29,7 @@ target 'VLC-iOS' do
   target 'VLC for iOSUITests' do
     pod 'SimulatorStatusMagic'
   end
+  target 'VLC for iOSTests'
```
Mikanbu: I'm not a 100% sure but I think we have three warnings with this addition(for example a the end of `pod install`).

Sources/URLHandler.swift
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    func performOpen(url: URL, options: [UIApplicationOpenURLOptionsKey: AnyObject]) -> Bool {
+        play(url: url, completion: nil)
+        return true
+    }
+}
+
+// TODO: This code should probably not live here
+func play(url: URL, completion: ((Bool) -> Void)?) {
```
Mikanbu: I realise that there is a `TODO` for it  but I really think we should encapsulated them even temporarily.
Form what I understand the two methods are only used by classes that conforms to `VLCURLHandler`.
What would you think adding them inside an extension of  `VLCURLHandler` like so: 


```extension VLCURLHandler {
    // TODO: This code should probably not live here
    func play(url: URL, completion: ((Bool) -> Void)?) {
      ...
    }

    func downloadMovie(from url: URL, fileNameOfMedia fileName: String?) {
        ...
    }
}
```

On the other hand, the actual call of the methods would have to be changed a tiny bit.

PR #111, opened 2018-08-06
VLCKeychainCoordinator: Fix that a playing video is not visible after…

carolanitz:
… dismissing the passcodecontroller

After coming back into the app from background and entering a password we had two titlebars.
Now we check for nil and superviews to avoid doubling (closes #157)

### Checklist


Mikanbu:
Merged with ded60c40c241d1e25f7a34fd59072515750092a1

PR #112, opened 2018-08-08
VLCMediaDatasource: after reordering just set the folderTrackNumber t…

carolanitz:
### Checklist

### DIFF
--- a/Sources/VLCMediaDataSource.m
+++ b/Sources/VLCMediaDataSource.m
@@ -89,11 +89,9 @@
             return;
         [_foundMedia removeObjectAtIndex:fromIdx];
         [_foundMedia insertObject:object atIndex:toIdx];
-        object.folderTrackNumber = @(toIdx - 1);
-        object = [_foundMedia objectAtIndex:fromIdx];
-        if (![object isKindOfClass:[MLFile class]])
-            return;
-        object.folderTrackNumber = @(fromIdx - 1);
+        for (MLFile *object in _foundMedia) {
+            object.folderTrackNumber = @([_foundMedia indexOfObject:object]);
+        }
     }
 }
 


Sources/VLCMediaDataSource.m
```diff
@@ -89,11 +89,9 @@ - (void)moveObjectFromIndex:(NSUInteger)fromIdx toIndex:(NSUInteger)toIdx
...
         [_foundMedia removeObjectAtIndex:fromIdx];
         [_foundMedia insertObject:object atIndex:toIdx];
-        object.folderTrackNumber = @(toIdx - 1);
-        object = [_foundMedia objectAtIndex:fromIdx];
-        if (![object isKindOfClass:[MLFile class]])
-            return;
-        object.folderTrackNumber = @(fromIdx - 1);
+        for (MLFile *object in _foundMedia) {
+            object.folderTrackNumber = @([_foundMedia indexOfObject:object]);
+        }
```
dcordero: Since `indexOfObject` requires a linear search, this method has probably a complexity of O(n), that together with the wrapping for loop means having a quadratic complexity.

If I understand this code correctly, I think we could improve the performance of this piece of code to be linear just using a counter to get the current index iteration without the need of searching.

```
var count = 0;
for (MLFile *object in _foundMedia) {
  object.folderTrackNumber = @(count);
  count++;
}
```

carolanitz: didn't even think of that 
thanks @dcordero !

carolanitz:
closed by af7e4edf7e4a2edcf0ad33458e11b3c530264a56 & b4ade073cb12c16d4d6c85b2bec3a986ed0a971c

PR #120, opened 2018-08-18
Fix running tests in CI build

alexaubry:
### Checklist

### Description

Tests were failing to build on CI, because Xcode could not find a class from the main app when building the test suite.

It seemed to be related to the contents of the derived data. These changes fixed the issue:

- Change the name of the module to “VLC”, so that we call `import VLC` in the test files
- Make the tests targets inherit the Pods search paths from the main app
- Use `VLC-iOS<Tests|UITests>` as the name of the test modules, so that they have a common prefix with the main app

Edits to the Fastfile were also made:

- Build the app for testing, so that we don’t need to build it a second time when running the tests (this seems to contribute to the failure in some proportion)
- Do not refresh the Pods repo when running the CocoaPods phase, as Circle already does this when bootstrapping the tests


alexaubry:
This issue seems to happen only when building from scratch. If you run `bundle exec fastlane ci` twice, it will fail the first time and succeed the second time. Might be related to derived data.

carolanitz:
thank you so so much \o/ I used a couple of your changes but didn't merge the fastfile changes. because I was a bit unsure if they were really necessary. If you think those alone should still be merged can you open a PR just with that?  merci ❤️ closing and partly merged with f03a9375aff2b18e2dedd7d2b885a68b30da3ade and a8ea7899753688e9590a5beff7b951f202424c29

PR #125, opened 2018-08-22
Fix warnings

dcordero:
### Checklist

### Description
Fix a few warnings here and there. 

Most of them because of referencing a private variable from a block, I changed them by private properties which is a quite common point in most of the objc style guides (i.e. https://github.com/raywenderlich/objective-c-style-guide#variables)

### DIFF
--- a/Apple-TV/VLCAboutViewController.m
+++ b/Apple-TV/VLCAboutViewController.m
@@ -65,9 +65,8 @@
     [self stopAnimation];
 }
 
-- (UIView *)preferredFocusedView
-{
-    return self.blablaTextView;
+- (NSArray<id<UIFocusEnvironment>> *)preferredFocusEnvironments {
+    return @[self.blablaTextView];
 }
 
 - (void)scrollToTop
--- a/README.md
+++ b/README.md
@@ -33,9 +33,10 @@ It's currently written in Objective-C / Swift and uses [VLCKit](https://code.vid
 
 ### VLC-iOS
 
-1. Run `pod install`.
-2. Open `VLC.xcworkspace`.
-3. Hit "Build and Run".
+1. Run `bundle install`.
+2. Run `bundle exec pod install`.
+3. Open `VLC.xcworkspace`.
+4. Hit "Build and Run".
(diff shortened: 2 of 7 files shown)


Apple-TV/VLCAboutViewController.m
```diff
@@ -65,9 +65,8 @@ - (void) viewDidDisappear:(BOOL)animated
     [self stopAnimation];
 }
 
-- (UIView *)preferredFocusedView
-{
-    return self.blablaTextView;
+- (NSArray<id<UIFocusEnvironment>> *)preferredFocusEnvironments {
+    return @[self.blablaTextView];
```
carolanitz: wow blablaTextView ??

README.md
```diff
@@ -33,9 +33,10 @@ It's currently written in Objective-C / Swift and uses [VLCKit](https://code.vid
 
 ### VLC-iOS
 
-1. Run `pod install`.
-2. Open `VLC.xcworkspace`.
-3. Hit "Build and Run".
+1. Run `bundle install`.
+2. Run `bundle exec pod install`.
```
carolanitz: yeah I forgot that this is better than just pod install can you explain again why though?

dcordero: Sure...

The project has some Ruby dependencies: fastlane, cocoapods, rubocop, xcode-install and xcodeproj. 

[Bundler](https://bundler.io) guarantees a consistent environment so that everybody (including CI) use the same versions of those dependencies. 

• `bundle install` installs all the dependencies on their expected versions. As defined by [Gemfile](https://code.videolan.org/videolan/vlc-ios/blob/master/Gemfile) and [Gemfile.lock](https://code.videolan.org/videolan/vlc-ios/blob/master/Gemfile.lock). (both already defined in the project)
• `bundle exec [command]` runs the command in a consistent environment with those dependencies installed.

Apple-TV/VLCAboutViewController.m
dcordero: Good point, this might need indeed a renaming.

PR #128, opened 2018-08-27
Actionsheet: Add theme observer (fixes #301)

mkchoi212:
### Checklist

### Description

Theme observer code for VLCActionSheet's has been moved inside `VLCActionSheet.swift` to keep things contained.

Bug reference https://code.videolan.org/videolan/vlc-ios/issues/301

### DIFF
--- a/Sources/VLCActionSheet/VLCActionSheet.swift
+++ b/Sources/VLCActionSheet/VLCActionSheet.swift
@@ -171,6 +171,8 @@ class VLCActionSheet: UIViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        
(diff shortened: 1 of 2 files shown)


Mikanbu:
IIRC, I put the theme code inside the `VLCRendererDiscovererManager` because the theme was related it. Meaning that since it is `VLCRendererDiscovererManager` that creates the action sheet,  it has the control over the sheet.
I don't mind this change, I can see the advantages, but wouldn't we lose the ability to customize it by component?

mkchoi212:
Aren't all the actionsheets styled the same way? Or did I miss something in Figma :p

carolanitz:
I believe this is the right change. As long as we don't need the ability to customize it by component somwhere we avoid this way that someone who uses this sheet has to think of this.

carolanitz:
merged with f44b4a9c93d0e5311936957b106ce9341f6bc903

PR #129, opened 2018-08-27
Actionsheet: Set rounded corners

mkchoi212:
### Checklist

### Description

Corner radius on top of the header has been applied, based on Figma mockups.

simulator screen shot - iphone x - 2018-08-26 at 21 45 46

### DIFF
--- a/Sources/VLCActionSheet/VLCActionSheet.swift
+++ b/Sources/VLCActionSheet/VLCActionSheet.swift
@@ -200,6 +200,12 @@ class VLCActionSheet: UIViewController {
         // This is to avoid a horrible visual glitch!
         mainStackView.isHidden = false
 
+        let roundedCornerPath = UIBezierPath(roundedRect: headerView.bounds, byRoundingCorners: [.topLeft, .topRight],
+                                             cornerRadii: CGSize(width: 10, height: 10))
+        let maskLayer = CAShapeLayer()
+        maskLayer.path = roundedCornerPath.cgPath
+        headerView.layer.mask = maskLayer
+
         UIView.transition(with: backgroundView, duration: 0.2, options: .transitionCrossDissolve, animations: { [weak self] in
             self?.backgroundView.isHidden = false
             }, completion: nil)


Sources/VLCActionSheet/VLCActionSheet.swift
```diff
@@ -200,6 +200,12 @@ class VLCActionSheet: UIViewController {
         // This is to avoid a horrible visual glitch!
         mainStackView.isHidden = false
 
+        let roundedCornerPath = UIBezierPath(roundedRect: headerView.bounds, byRoundingCorners: [.topLeft, .topRight],
+                                             cornerRadii: CGSize(width: 10, height: 10))
+        let maskLayer = CAShapeLayer()
+        maskLayer.path = roundedCornerPath.cgPath
+        headerView.layer.mask = maskLayer
```
carolanitz: you can just set the cornerRadius of the views layer

mkchoi212: Just tried it but didn't work :(

The actionSheet is `headerView` + `stackView` and so I think the corner radius of the headerView must be set.

carolanitz: you also need to set maskstobounds I think

mkchoi212: Yep tried that too 🤭

mkchoi212: Ah I should have been clearer! Sorry about that

carolanitz: yeah it worked with just setting the cornerradius of the headerview but it obviously didn't have the wanted effect. I'm actually curious how the Maps app build it

mkchoi212: Hmmm good question. I don’t think they are using a stack view for the sheet unlike us. Maybe just a table view with the table’s corner radius set??

carolanitz:
This was merged with 99d76f24cf7f56179e137d9d6c4041897c2a0b38

PR #130, opened 2018-08-27
Move VLCMediaFileDiscovererDelegate to VLCMediaLibraryManager

Mikanbu:
### Checklist
- [x] I've read the [Contribution Guidelines](https://gfrithub.com/videolan/vlc-ios#contribute)

### Description

This allows monitoring of changes in the Documents files from external
apps.

Additionally changes `mediaFilesFoundRequiringAdditionToStorageBackend`
to optional since the new medialibrary doesn't need it.

### DIFF
--- a/SharedSources/VLCMediaLibraryManager.swift
+++ b/SharedSources/VLCMediaLibraryManager.swift
@@ -88,11 +88,20 @@ class VLCMediaLibraryManager: NSObject {
 
     // MARK: Private
 
+    private func setupMediaDiscovery(at path: String) {
+        let mediaFileDiscoverer = VLCMediaFileDiscoverer.sharedInstance()
+        mediaFileDiscoverer?.directoryPath = path
+        mediaFileDiscoverer?.addObserver(self)
+        mediaFileDiscoverer?.startDiscovering()
+    }
+
     private func setupMediaLibrary() {
         guard let documentPath = NSSearchPathForDirectoriesInDomains(.documentDirectory, .userDomainMask, true).first,
             let libraryPath = NSSearchPathForDirectoriesInDomains(.libraryDirectory, .userDomainMask, true).first else {
                 preconditionFailure("VLCMediaLibraryManager: Unable to init medialibrary.")
         }
+
+        setupMediaDiscovery(at: documentPath)
 
         let databasePath = libraryPath + "/MediaLibrary/" + VLCMediaLibraryManager.databaseName
         let thumbnailPath = libraryPath + "/MediaLibrary/Thumbnails"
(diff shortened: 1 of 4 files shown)


Sources/VLCAppDelegate.m
```diff
@@ -283,36 +277,6 @@ - (void)application:(UIApplication *)application performActionForShortcutItem:(U
...
-#pragma mark - media discovering
-
-- (void)mediaFileAdded:(NSString *)fileName loading:(BOOL)isLoading
-{
-    if (!isLoading) {
-        MLMediaLibrary *sharedLibrary = [MLMediaLibrary sharedMediaLibrary];
-        [sharedLibrary addFilePaths:@[fileName]];
-
-        /* exclude media files from backup (QA1719) */
-        NSURL *excludeURL = [NSURL fileURLWithPath:fileName];
```
carolanitz: can you make sure that we exclude mediafiles in the future also from backup?

Mikanbu: By mediafiles do you mean all the users media?

carolanitz: yes

Mikanbu: Updated!

SharedSources/VLCMediaLibraryManager.swift
```diff
@@ -219,6 +228,26 @@ extension VLCMediaLibraryManager {
...
+    func mediaFileAdded(_ filePath: String!, loading isLoading: Bool) {
+        guard !isLoading else {
+            return
+        }
+        /* exclude media files from backup (QA1719) */
+        var excludeURL = URL(fileURLWithPath: filePath)
+        var resourceValue = URLResourceValues()
+
+        resourceValue.isExcludedFromBackup = true
+        try? excludeURL.setResourceValues(resourceValue)
```
carolanitz: can you catch the failure here ?

Sources/VLCMediaFileDiscoverer.h
```diff
@@ -14,10 +14,8 @@
 
 @protocol VLCMediaFileDiscovererDelegate <NSObject>
 
-@required
-- (void)mediaFilesFoundRequiringAdditionToStorageBackend:(NSArray <NSString *> *)foundFiles;
-
 @optional
+- (void)mediaFilesFoundRequiringAdditionToStorageBackend:(NSArray <NSString *> *)foundFiles;
```
carolanitz: why did you move it to optional and do we still need it if it's not implemented?

SharedSources/VLCMediaLibraryManager.swift
```diff
@@ -219,6 +228,26 @@ extension VLCMediaLibraryManager {
...
+
+        reload()
+    }
+
+    func mediaFileDeleted(_ filePath: String!) {
+        reload()
```
carolanitz: We should not have an entire reload for mediaFileDeleted. Instead we should just remove from the collectionView the file that belongs to that path

carolanitz:
there are still commented out references in the Appdelegate

Mikanbu:
> there are still commented out references in the Appdelegate

Are you talking about l268? If so, it's unrelated to this PR, it should be addressed in the migration one since it touches the old migration code.

As for removing the use of `VLCMediafileDiscoverer` inside `VLCDocumentPickerController`, I don't think it should be addressed in this PR as well.

Sources/VLCMediaFileDiscoverer.h
Mikanbu: As I can see it's used in tvOS and can't say if it's really needed but I really don't want to start refactoring tvOS related code here. 

Additionally, my goal for this PR is not the remove the usage of `VLCMediaFileDiscoverer`.

For the why I removed it, as written in the commit description, I changed it to optional since the new medialibrary doesn't need a callback for `mediaFilesFoundRequiringAdditionToStorageBackend`.

SharedSources/VLCMediaLibraryManager.swift
Mikanbu: Here, I'm reloading the medialibrary not the UI.
UI related updates(addition/deletion by indexes) should be addressed in another PR in the respective models.

SharedSources/VLCMediaLibraryManager.swift
Mikanbu: Will do for the failure catching!

carolanitz:
> Are you talking about l268? If so, it's unrelated to this PR, it should be addressed in the migration one since it touches the old migration code.

yes l268 but there is also still a call in the migration completion. For me it would've made sense since you're moving the entire Mediafilediscoverer but that's fine

PR #136, opened 2018-08-30
VLCMovieViewController: adjust activityIndicator behavior

carolanitz:
* position indicator in the middle of the movieViewController
* simplify state code and only show the indicator for buffering when we were in playing mode before
since we never get an event that we're done with buffering (closes #306)

### Checklist


carolanitz:
Now it shows when it buffers for chromecast but doesn't when you actually hit pause and seek.
Ideally, we should propagate the percentage value that comes with the buffering event and check if it is 100% and then not show the activityIndicator

PR #137, opened 2018-09-03
MediaCollectionViewCell: Added a basic MediaCell and structure for ad…

carolanitz:
…ding more cells

This also removes the old VLCPlaylistTableViewCell and VLCPlaylistCollectionViewCell

### Checklist

### Description

 Issues that still need to be addressed:
-  [ ] The cell is not dynamic in height, it will be the same height for titles with one line and titles with two lines
-  [ ] The cells don't respect the notch 
-  [ ] The cells don't update yet when the media updates. This is because the code doesn't update media at this point when playback is done and the file was played


SharedSources/MovieCollectionViewCell.swift
```diff
@@ -0,0 +1,64 @@
...
+    @IBOutlet weak var descriptionLabel: UILabel!
+    @IBOutlet weak var progressView: UIProgressView!
+
+    override class var nibName: String {
+        return "MovieCollectionViewCell"
+    }
+
+    override var media: VLCMLObject? {
+        didSet {
+            guard let movie = media as? VLCMLMedia else {
```
carolanitz: all this code should move into a function

SharedSources/MovieCollectionViewCell.swift
```diff
@@ -0,0 +1,64 @@
...
+        backgroundColor = PresentationTheme.current.colors.background
+        titleLabel?.textColor = PresentationTheme.current.colors.cellTextColor
+        descriptionLabel?.textColor = PresentationTheme.current.colors.cellDetailTextColor
+    }
+
+    override func prepareForReuse() {
+        super.prepareForReuse()
+        titleLabel.text = ""
+        descriptionLabel.text = ""
+        thumbnailView.image = nil
```
carolanitz: progress isn't handled here and the newLabel

SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
```diff
@@ -44,6 +45,10 @@ protocol MLBaseModel: MediaLibraryBaseModel {
 }
 
 extension MLBaseModel {
+    var cellType: MediaCollectionViewCell.Type {
+        return MovieCollectionViewCell.self
```
Mikanbu: I'm not sure to understand this addition, what happens will with the GenreModel for example?

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -224,7 +223,8 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
         var cellWidth = collectionView.bounds.size.width / numberOfCells
         cellWidth = cellWidth - ceil(((numberOfCells + 1) * cellPadding) / numberOfCells)
 
-        return CGSize(width: cellWidth, height: cellWidth * aspectRatio)
+        // 3*20 for the labels + 24 for 3*8 which is the padding
+        return CGSize(width: cellWidth, height: cellWidth * aspectRatio + 3*20+24)
```
carolanitz: this part will be adjusted with a dynamic calculation

SharedSources/MediaLibraryModel/VideoModel.swift
```diff
@@ -95,12 +95,25 @@ extension VLCMLMedia {
...
     }
 
     @objc func formatSize() -> String {
         return ByteCountFormatter.string(fromByteCount: Int64(mainFile().size()),
                                          countStyle: .file)
     }
+
+    func mediaProgress() -> Float {
+        guard let string = metadata(of: .progress).str as NSString? else {
+            return 0.0
```
Mikanbu: Is this for safety?

Sources/VLCPlaybackController.m
```diff
@@ -375,9 +375,7 @@ - (void)_savePlaybackState
         APLog(@"failed to save current media state - file removed?");
     }
 }
-#endif
 
-#if TARGET_OS_IOS
```
Mikanbu: I don't understand this VLCPlaybackController change in this commit.
I believe it should be a separate commit.

SharedSources/MovieCollectionViewCell.swift
```diff
@@ -0,0 +1,64 @@
...
+class MovieCollectionViewCell: MediaCollectionViewCell {
+
+    @IBOutlet weak var thumbnailView: UIImageView!
+    @IBOutlet weak var titleLabel: UILabel!
+    @IBOutlet weak var newLabel: UILabel!
+    @IBOutlet weak var descriptionLabel: UILabel!
+    @IBOutlet weak var progressView: UIProgressView!
+
+    override class var nibName: String {
+        return "MovieCollectionViewCell"
```
Mikanbu: What about using `String(describing:)` here?

SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
carolanitz: It's a fallback for now until this is implemented

SharedSources/MediaLibraryModel/VideoModel.swift
carolanitz: when there is no progress yet set the default will be 0

Sources/VLCPlaybackController.m
carolanitz: could be. I just saw it and removed it

SharedSources/MovieCollectionViewCell.swift
carolanitz: could use it yes

SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
carolanitz: exactly :)

Mikanbu: Since it won't always be a media, I think we should we consider renaming it so that it can be more generic.

What about`BaseCollectionViewCell`?

carolanitz: oh interesting so I can't always set a media on the cell? so I guess the base type is a VLCMLFile ?

Mikanbu: >  I can't always set a media on the cell?

Not sure to understand what you mean here, for example, you cannot have `VLCMLMedia` in a genre category.

The base class of the VLCMediaLibraryKit models would be `VLCMLObject`.
`VLCMLFile` is different from the old MediaLibrary, it is files contained in a certain Media.

Mikanbu:
Merged with d5225e9b522d22c0f057606c619c90ae5677fdb1.

PR #138, opened 2018-09-03
Remove Userwarnings: These warnings have no effect

carolanitz:
### Checklist

### Description

In development everyone will always work without the credentials and these warnings have therefore no value. If we want to ensure that we are not able to build without the credentials we could add some kind of assertion that fails if we build for production only when the keys are not set


carolanitz:
merged with 1b0a21b88e6d2d28c2e271394ccbea534045658c

PR #139, opened 2018-09-05
VLCLibraryViewController: Make reordering possible only in a folder

Mikanbu:
### Checklist

### Description

closes [#313](https://code.videolan.org/videolan/vlc-ios/issues/313)

Tested this change on an iPad as well for the drag and drop, this doesn't seem to affect it.
I may have missed some DnD cases.


carolanitz:
can you double check that this doesn't break drag and drop?

Mikanbu:
Already did as stated in the description~

PR #140, opened 2018-09-05
MediaEditCell: adjusted the cell to be up to spec

carolanitz:
### Checklist

### Description

this just adjusts the editcell to look like this: 
screen shot 2018-09-05 at 5 19 58 pm


carolanitz:
but Caro the padding between the title and the description is not enough ... duh 🙈yeah I adjust it

Resources/en.lproj/Localizable.strings
```diff
@@ -51,10 +51,8 @@
 "CONTINUE_PLAYBACK" = "Continue playback?";
 "CONTINUE_PLAYBACK_LONG" = "Would you like to continue playback of \"%@\" where you left off?";
 
-"BUTTON_ALL" = "Select All";
```
carolanitz: but Caro why the heck did you check in the Localizables .... yeah yeah I'm gonna remove that

Mikanbu:
Didn't tested fully but it seems that the first selection of a cell doesn't check the circle 🤔

carolanitz:
You can, but if you click directly on the checkmark it will receive the touch and not forward it to the cell so I need to make it userinteraction enabled = false

carolanitz:
I was wrong, I mistakenly showed the wrong image for checked and unchecked. this should be fixed now

SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
```diff
@@ -49,6 +50,10 @@ extension MLBaseModel {
         return MovieCollectionViewCell.self
     }
 
+    var editCellType: BaseCollectionViewCell.Type {
```
Mikanbu: I'm not sure here but do we really need this property inside `MediaLibraryBaseModel`? I'm wondering if this is the right place/right level of abstraction.

Sources/VLCEditController.swift
```diff
@@ -18,37 +18,36 @@ class VLCEditController: NSObject {
...
     private let category: MediaLibraryBaseModel
 
-//    private lazy var editToolbar: VLCEditToolbar = {
-//        let editToolbar = VLCEditToolbar(frame: CGRect(x: 0, y: 400,
-//                                                       width: collectionView.frame.width, height: 50))
-//        editToolbar.isHidden = true
-//        editToolbar.delegate = self
-//        return editToolbar
-//    }()
+    private lazy var editToolbar: VLCEditToolbar = {
```
Mikanbu: Out of curiosity, shouldn't we leave this commented out since we don't yet have a toolbar?

SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
carolanitz: I wouldn't know where else to be honest since the UI will be different for albums and playlists I guess

Sources/VLCEditController.swift
carolanitz: yeah can do

PR #141, opened 2018-09-06
PlaybackSettings: persist changes like playbackspeed, repeat, shuffle…

carolanitz:
… and aspectratio between playback

This also removes the user setting for the default playbackspeed since this is not needed anymore

### Checklist

### DIFF
--- a/Apple-TV/PlaybackSettings.swift
+++ b/Apple-TV/PlaybackSettings.swift
@@ -1,0 +1,73 @@
+/*****************************************************************************
+ * VLC for iOS
+ *****************************************************************************
+ * Copyright (c) 2018 VideoLAN. All rights reserved.
+ * $Id$
+ *
+ * Authors: Carola Nitz <caro # videolan.org>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import Foundation
+
+@objcMembers
+class PlaybackSettings: NSObject, Codable {
+    var repeatMode: VLCRepeatMode
+    var shuffleMode: Bool
+    var playbackSpeed: CGFloat
+    var aspectRatio: VLCAspectRatio
+
+    private enum CodingKeys: String, CodingKey {
+        case repeatMode
+        case shuffleMode
+        case playbackSpeed
+        case aspectRatio
+    }
+    
+    override init() {
+        repeatMode = .doNotRepeat
+        shuffleMode = false
+        aspectRatio = .default
+        let defaults = UserDefaults.standard
+        if defaults.float(forKey: "playback-speed") != 0 {
+            playbackSpeed = CGFloat(defaults.float(forKey:"playback-speed"))
+        } else {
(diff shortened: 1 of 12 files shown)


Apple-TV/PlaybackSettings.swift
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * Authors: Carola Nitz <caro # videolan.org>
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import Foundation
+
+@objcMembers
+class PlaybackSettings: NSObject, Codable {
```
Mikanbu: In this case since it's a data structure, what do you think about using a structure?

Mikanbu: Would you mind explaining the choice of making this Codable with the choice of JSON representation ?

Apple-TV/PlaybackSettings.swift
```diff
@@ -0,0 +1,73 @@
+/*****************************************************************************
```
Mikanbu: I think this file shouldn't be inside Apple-TV folder.
Perhaps it should be inside SharedSources?

Sources/VLCPlaybackController.m
```diff
@@ -793,12 +784,17 @@ - (void)pause
     [[NSNotificationCenter defaultCenter] postNotificationName:VLCPlaybackControllerPlaybackDidPause object:self];
 }
 
+- (void)setShuffleMode:(BOOL)shuffleMode
+{
+    _shuffleMode = _pbcSettings.shuffleMode = shuffleMode;
```
Mikanbu: Do we still need to keep a copy the different playback setting properties? Could we expose `PlaybackSettings` directly?

Apple-TV/PlaybackSettings.swift
carolanitz: yes that way I only need to store the playbacksettings as one object and not every single property

Apple-TV/PlaybackSettings.swift
carolanitz: true, But I think it is about time that we add swift to AppleTV

Sources/VLCPlaybackController.m
carolanitz: Ah you mean we should remove those settings entirely from the playbackcontroller ? yes I like that idea

carolanitz: You need to set this information on the _listplayer or _mediaplayer. The settings are just holding that information. I would need to have a settingsDidChange method.

carolanitz:
I will remove the aspectRatio though because it is media specific

carolanitz:
This shouldn't be merged until we uniformly agree on the handling on states across all players

carolanitz:
what we agreed on has been stated here: https://wiki.videolan.org/Playback_States/

PR #142, opened 2018-09-06
MediaLibraryModel: Refactor to MediaModel

Mikanbu:
### Checklist

### Description

Add a level of abstraction for AudioModel and VideoModel.
This removes some duplication and will help later on mostly on the edit part.


SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
```diff
@@ -25,7 +25,7 @@ protocol MediaLibraryBaseModel {
     func sort(by criteria: VLCMLSortingCriteria)
 }
 
-protocol MLBaseModel: MediaLibraryBaseModel {
+protocol MLBaseModel: class, MediaLibraryBaseModel {
```
carolanitz: never seen this, what does this do ?

Mikanbu: It makes it a class only protocol, it's the same thing as AnyObject but actually I will change it to AnyObject. 😄

carolanitz: yeah I actually read up on it

PR #143, opened 2018-09-10
Fix overlapping title

carolanitz:
### Checklist

### Description

This moves the PlayingExternally layout into one xib, this has the advantage that we can use autolayout at least for that view and don't, in the future, update the screen in the iPhone version and forget the iPad xib. It also displays the movie title now as playing externally title which doesn't lead to colliding text.


carolanitz:
yes and I should move the xib in a different subfolder

PR #145, opened 2018-09-14
AppDelegate: Change viewController hierachy to fix wrong frame of min…

carolanitz:
…iplayerView

When you add a childViewConroller to a UITabbarcontroller it will keep them in the viewControllers.
by setting the viewControllers on the tabbarcontroller later we removed the childViewController which led to that Controller
not getting any updates like willTransitionToSize. Simply appending the other viewcontrollers would show the player as additional tab
We now have plain UiViewController with tabbarController and PlayerController as ChildViewControllers to fix this layout issue

### Checklist

### Description

 
I still need to figure out how to animate the transition alongside the screen rotation but this brings us a step further

### DIFF
--- a/Sources/Coordinators/AppCoordinator.swift
+++ b/Sources/Coordinators/AppCoordinator.swift
@@ -18,29 +18,37 @@ class Services: NSObject {
 
 @objc class AppCoordinator: NSObject {
     private var childCoordinators: [NSObject] = []
-    private var tabBarController: UITabBarController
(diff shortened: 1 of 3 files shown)


Mikanbu:
I'm sorry, after reading the description I'm a bit confused by what this PR is trying to resolve/the why we would want it?

carolanitz:
merged with 1b4747e4a37332d4542869a217fb23c93962bcd3

PR #146, opened 2018-09-14
Add audio cell

carolanitz:
### Checklist

### Description

Thumbnails are still missing for Audio tracks
simulator screen shot - iphone 8 plus - 2018-09-14 at 20 06 28
simulator screen shot - iphone 8 plus - 2018-09-14 at 20 06 23


MediaCategoryCells/BaseCollectionViewCell.swift
```diff
@@ -23,4 +23,12 @@ class BaseCollectionViewCell: UICollectionViewCell {
     }
 
     var media: VLCMLObject?
+
+    class func cellSizeForWidth(_ width: CGFloat) -> CGSize {
+        return CGSize.zero
+    }
+
+    class var cellPadding: CGFloat {
```
Mikanbu: Small cosmetic thing but properties should be before methods.

SharedSources/MediaLibraryModel/AlbumModel.swift
```diff
@@ -16,6 +16,8 @@ class AlbumModel: MLBaseModel {
 
     var files = [VLCMLAlbum]()
 
+    var cellType: BaseCollectionViewCell.Type { return AudioCollectionViewCell.self }
```
Mikanbu: Should we consider putting another cell while the specific cell is done? It can be a pain if the app fatal Errors each time you want to change tab that doesn't have an `VLCMLMedia`.

Mikanbu:
Is this the desired behaviour in landscape? 

img_5707

Mikanbu:
Should we consider a ticket for dynamic cell sizes?

MediaCategoryCells/AudioCollectionViewCell.swift
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+
+    override func awakeFromNib() {
+        super.awakeFromNib()
+        newLabel.text = NSLocalizedString("NEW", comment: "")
+        newLabel.textColor = PresentationTheme.current.colors.orangeUI
+        NotificationCenter.default.addObserver(self, selector: #selector(themeDidChange), name: .VLCThemeDidChangeNotification, object: nil)
+        themeDidChange()
+        thumbnailView.layer.cornerRadius = thumbnailView.frame.size.width / 2.0
```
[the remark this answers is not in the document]
carolanitz: I don't think you're able to set it in reference to the thumbnailframe size in a xib

carolanitz:
> Is this the desired behaviour in landscape?
> 
> img_5707

this design was edited with the dynamic sizing and looks different when you build it

carolanitz:
> Should we consider a ticket for dynamic cell sizes?

that was done already as part of the second commit ;)

SharedSources/MediaLibraryModel/AlbumModel.swift
carolanitz: very true I can change that

carolanitz: I just put moviewcell everywhere. I'll update the ticket

PR #148, opened 2018-09-19
Use only DWARF format in debug

Mikanbu:
### Checklist

### Description

This change the build settings changing the debug information format.
With this PR, `DWARF with dSYM` is used only in distribution and release.

The why of this change is mostly:

 - Speed up "compilation" for debug.
 - Doesn't make my mac freeze for 15-ish minutes with Xcode 10(https://forums.developer.apple.com/message/330343#330343).

I realise this is mostly a workaround for the Xcode 10 issue but nonetheless, I don't think we need dSYM file in debug.


carolanitz:
yeah I was actually looking into that as well and think it's a good idea!

carolanitz:
merged with 4269d8f759aca8023c107c130b1ac68f375838ee

PR #150, opened 2018-10-01
PlaylistViewController: remove separate video and audio playlists and…

carolanitz:
… add playlist tab

### Checklist

### Description

as discussed at VDD

### DIFF
--- a/Sources/MediaViewControllers/PlaylistViewController.swift
+++ b/Sources/MediaViewControllers/PlaylistViewController.swift
@@ -1,5 +1,5 @@
 /*****************************************************************************
- * VideoViewController.swift
+ * VLCPlaylistViewController.swift
  * VLC for iOS
  *****************************************************************************
  * Copyright (c) 2018 VideoLAN. All rights reserved.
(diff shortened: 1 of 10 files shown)


Sources/MediaViewControllers/PlaylistViewController.swift
```diff
@@ -0,0 +1,33 @@
...
+        tabBarItem = UITabBarItem(
+            title: NSLocalizedString("PLAYLISTS", comment: ""),
+            image: UIImage(named: "TVShowsIcon"),
+            selectedImage: UIImage(named: "TVShowsIcon"))
+        tabBarItem.accessibilityIdentifier = VLCAccessibilityIdentifier.playlist
+    }
+
+    override func viewControllers(for pagerTabStripController: PagerTabStripViewController) -> [UIViewController] {
+        return [
+            VLCPlaylistCategoryViewController(services)
```
Mikanbu: Is there a way to disable `PagerTabStrip...` when there is only one VC, I think it will look kind of odd with one item?

carolanitz: I actually wouldn't want to do that right now and discuss it with Louis. Right now it looks alike and in sync with the other controllers

Mikanbu: I'm sorry if I misunderstood but, are you saying we should merge and see later on with Louis or we should freeze this PR until Louis input?

I agree that his input is important but on some cases I think we can be independent.
I believe this case is one of them mainly due to:

 - Redundancy of Playlists "label" on top
 - Scroll on sides
 - Unnecessary usage of top space for a single item

gifplaylist

I'm always open for discussion if you have other opinions on the matter.

carolanitz: The bouncing should definitely be disabled but I wouldn't want to remove the top above. I'd rather just change the text.  I can see already issues in regards to editbar positioning or emptyview positioning which is done the same on video, audio and playlistviewcontrollers and the little bit of space that we loose is okay if the behavior and UI is consistent
And yes besides the bouncing I would address the rest in a follow up commit

carolanitz:
@Mikanbu ping?

carolanitz:
updated it to remove the double playlist

carolanitz:
simulator screen shot - iphone xr - 2019-01-31 at 15 55 09

carolanitz:
merged with 849de3812263aa531546f3e62ef852801c82b709

PR #154, opened 2018-10-10
Fix chapter descriptions on tvOS

floatingatoll:
### Checklist

### Description
A tvOS bug exists where all chapters show the name of the first chapter, rather than the names of each chapter. This is because the player's current title's index, which has nothing to do with the chapter object being processed here, is used to lookup in the chapter descriptions array. Using the current chapter's index instead results in a valid lookup and working chapter titles on my development tvOS device.

Note that the fastlane test above does not exercise this controller, and so while the tests passed, I feel guilty checking that box without declaring the caveat. Duly noted.

```
[13:03:33]: ▸ Test Succeeded
+--------------------+----+
|      Test Results       |
+--------------------+----+
| Number of tests    | 13 |
| Number of failures | 0  |
+--------------------+----+
```

### DIFF
--- a/Apple-TV/Playback/Playback Info/VLCPlaybackInfoChaptersTVViewController.m
+++ b/Apple-TV/Playback/Playback Info/VLCPlaybackInfoChaptersTVViewController.m
@@ -135,7 +135,7 @@
     trackCell.selectionMarkerVisible = isSelected;
 
     VLCPlaybackController *vpc = [VLCPlaybackController sharedInstance];
-    NSDictionary *description = [vpc chapterDescriptionsDictAtIndex:[vpc indexOfCurrentTitle]];
+    NSDictionary *description = [vpc chapterDescriptionsDictAtIndex:row];
 
     NSString *chapter = description[VLCChapterDescriptionName];
     if (chapter == nil)


floatingatoll:
Thanks! I could never have met all the criteria demanded for bug reports but I knew it was probably an easy fix. I hope it helps.

fkuehne:
You are welcome to add more bug reports or feature requests on https://code.videolan.org - is there something we can do to ease reporting for you?

fkuehne:
Merged with master and 3.0 branch.

floatingatoll:
I notice a lot of minor UX issues that don't really seem to be taken into account by the guidelines https://wiki.videolan.org/Report_bugs/ — and the "Information to send" section is more targeted towards decoder/media issues than on UX issues with the VLC product itself. So I've never felt like UX bugs would be welcome as a result, and instead have learned enough to submit pull requests :)

fkuehne:
Wow, I believe you are one of the first in years to read those guidelines so thoroughly! :)

We would be very happy to receive feedback on UX (as well as more pull requests!) and there is no fixed form to share them with us. Filing a ticket on https://code.videolan.org would be very welcome.

Note that for iOS, there is a major re-write of the UI under way, so it might be worth it to compile the master branch.

PR #155, opened 2018-10-10
Layout the subframe so that it's shown on first visit

floatingatoll:
### Checklist

### Description
On tvOS, there's a bug where the list of chapters isn't shown when the info pane is first opened. Switching to another tab of the pane and then back to the (default, first) Chapters tab displays the subview correctly. Calling layoutIfNeeded when binding the subview to the info controller seems to ensure that it's ready to be displayed on the first visit.

Note that the fastlane tests don't seem to test tvOS, but it does pass them, and this change _is_ effective on my local tvOS device — however I am not well-versed in the intricacies of layoutIfNeeded, so please use your best judgement here.


carolanitz:
merged with 6bb63a1f11714ee768b5a43194b47c1f0fa765bf

PR #157, opened 2018-10-13
Delay enabling passcode until after it has been set

segiddins:
### Checklist

### Description

The app now allows toggling the passcode setting to "enabled" and then killing/backgrounding the app. Before, it would (upon re-launch) assert that a passcode was in the keychain, even though the user would never have entered a passcode. Now, we enforce that passcode is not enabled until we have finished setting a passcode.


carolanitz:
merged with 81e93935e67d022a6db903443d873fadeb756525 & 6509de33fb65031f6e1dee734beec7ff7a4ca116 & 0b99e886910e6121af578ab1a01b173d89ae64f7

PR #158, opened 2018-10-13
Keep showing the video view when pausing a video in the mini player

segiddins:
### Checklist

### Description

Before, pausing a video via the mini player would completely hide the video thumbnail, and it would stay hidden even after re-starting the video. With this change, we hide the video view when playing audio, and show it otherwise.


carolanitz:
merged thank you :)

PR #167, opened 2018-11-13
VLCMovieViewControlPanel: adjust for iPad Pro to not be completely fl…

carolanitz:
…ush against the border

### Checklist

### Description

This pr makes sure that the volumeview and actionbutton are not flush with the edges but layout in respect to the layoutmargin on iPads


carolanitz:
the volumebar is more indented and has the same inset as the items of the statusbar as can be seen here in the before and after pictures
img_0025
img_0026

carolanitz:
merged with e82dc4272b9c6fe348557c6ad31b8296da3c6718

PR #176, opened 2018-11-22
Fix rendererButton size to 44x44

Mikanbu:
### Checklist

### DIFF
--- a/Sources/VLCRendererDiscovererManager.swift
+++ b/Sources/VLCRendererDiscovererManager.swift
@@ -132,7 +132,7 @@ class VLCRendererDiscovererManager: NSObject {
     ///
     /// - Returns: New `UIButton`
     @objc func setupRendererButton() -> UIButton {
-        let button = UIButton()
+        let button = UIButton(frame: CGRect(x: 0, y: 0, width: 44, height: 44))
         button.isHidden = getAllRenderers().isEmpty
         button.tintColor = PresentationTheme.current.colors.orangeUI
         button.setImage(UIImage(named: "renderer"), for: .normal)


Sources/VLCLibraryViewController.m
```diff
@@ -414,10 +417,14 @@ - (void)_displayEmptyLibraryViewIfNeeded
...
-        self.navigationItem.rightBarButtonItems = @[self.editButtonItem, _rendererBarButton];
+
+        UIBarButtonItem *fixedSpace = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemFixedSpace target:nil action:nil];
+        fixedSpace.width = NAVBAR_FIXED_SPACE;
+        self.navigationItem.rightBarButtonItems = @[self.editButtonItem, fixedSpace, _rendererBarButton];
         if (UI_USER_INTERFACE_IDIOM() == UIUserInterfaceIdiomPad) {
+            fixedSpace.width = NAVBAR_FIXED_SPACE_IPAD;
             self.displayModeBarButtonItem = [[UIBarButtonItem alloc] initWithImage:[UIImage imageNamed:@"tableViewIcon"] style:UIBarButtonItemStylePlain target:self action:@selector(toggleDisplayedView:)];
-            self.navigationItem.rightBarButtonItems = @[self.displayModeBarButtonItem, self.editButtonItem, _rendererBarButton];
+            self.navigationItem.rightBarButtonItems = @[self.displayModeBarButtonItem, self.editButtonItem, fixedSpace, _rendererBarButton];
```
carolanitz: how about using a stackview here btw ?

Mikanbu: Might be missing something but I don't think I can assign a stackView to an array here?

carolanitz: you can put a stackview in an array though

Mikanbu: I don’t see why.
Doesn’t it get handled by a stackview internally anyway?

carolanitz: I dislike the fixedspace. I'd rather a bigger button and a stackview would take care of that by scaling them all to the same size. That would increase the hittable area

carolanitz:
@Mikanbu I just understood why there is no padding in the first place. The renderer button is too small. Buttons by Apples UX guideline should have roughly 44x44 size, the renderer button is far from it. here is the reference (https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/adaptivity-and-layout/) if you fix that you won't need the fixed space. I believe that is an issue in 3.0 and master

Mikanbu:
Updated!

PR #190, opened 2019-01-30
VLCMediaLibraryManager: we have to wait until parsing of media is done

carolanitz:
### Checklist


carolanitz:
@Mikanbu no not the tests need fixing. The migration controller shows when he shouldn't which is why the tests are failing

Mikanbu:
>  The migration controller shows when it shouldn't...

From my understanding, since the test is run on a new device, the MediaLibraryManager will always start the migration.

carolanitz:
closing and follow up work is done with https://github.com/videolan/vlc-ios/pull/216. Had a messed up state since this was created with an old machine

PR #198, opened 2019-02-19
Medialibrary: Sort by category

Mikanbu:
### Checklist

### Description

This PR aims to refactor and integrate sorting by category.

Currently, the actionsheet doesn't take into account the dark mode since `UIAlertController` actionsSheet's cancel button can't be customized unless we go through private APIs.

The solution for this can be:

- Using our own action sheet, therefore, having a specific style.
- Create a XiB in order to achieve the result shown in this [SO answer](https://stackoverflow.com/a/53424527/5493755).

Additionally, I'm using a work arround in order to propagate `handleSort` onto child VC's. I would love to avoid this, so ideas are welcome! 😄


Mikanbu:
Updated PR to add related ticket [#162](https://code.videolan.org/videolan/vlc-ios/issues/162).

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -235,23 +205,67 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
...
+        var alertActions = [UIAlertAction]()
+
+        for (index, enabled) in category.sortModel.sortingCriteria.enumerated() {
+            guard enabled else { continue }
+            let criteria = VLCMLSortingCriteria(value: UInt(index))
+
+            alertActions.append(UIAlertAction(title: criteria.description, style: .default) {
+                [weak self] action in
+                self?.category.sortModel.currentSort = criteria
+                self?.category.sort(by: criteria)
```
carolanitz: How do you feel about changing the currentSort automatically triggers the sort? If you have the above you could theoretically set a currentSort to one thing and then sort by something else

Mikanbu: You're totally right, I believe this should be inside `sort()`.

PR #202, opened 2019-02-26
Edit

carolanitz:
### Checklist

### Description

Editing itself doesn't work yet. This just provides the UI for it and some stubs to implement. 
It also shows e.g. the rename button still for Artist, Album and Genre even though that option shouldn't be available in these Controllers

edit


carolanitz:
addresses big parts of https://code.videolan.org/videolan/vlc-ios/issues/187

Sources/VLCEditToolbar.swift
```diff
@@ -13,30 +13,89 @@ protocol VLCEditToolbarDelegate: class {
...
     func createPlaylist()
     func rename()
+    func share()
 }
 
-// Decided to use a UIView instead of UIToolbar because we have more freedom
-// FIXME: Basic structure without UI
 class VLCEditToolbar: UIView {
     weak var delegate: VLCEditToolbarDelegate?
+    private var category: MediaLibraryBaseModel
```
Mikanbu: I might be mistaken here, it seems that the category is only used for file type detection.
What would you think about using only the file type to avoid sending unused information?
Something like:

```Swift

private var fileType: VLCMLObject.Type

...

if !(fileType is VLCMLAlbum.Type) && !(file is VLCMLArtist.Type) && !(file is VLCMLGenre.Type) { ... }

```

Sources/VLCEditToolbar.swift
```diff
@@ -13,30 +13,89 @@ protocol VLCEditToolbarDelegate: class {
...
+        let deleteButton = UIButton(type: .system)
+        deleteButton.addTarget(self, action: #selector(deleteSelection), for: .touchUpInside)
+        deleteButton.setImage(UIImage(named: "delete"), for: .normal)
+        deleteButton.tintColor = .orange
+        deleteButton.widthAnchor.constraint(equalToConstant: 44).isActive = true
 
-    @objc func createFolder() {
+        return deleteButton
+    }()
+    private var addtoPlaylist: UIButton = {
```
Mikanbu: Small capitalization missing on `To`.

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -108,6 +117,19 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
...
+        editToolbar.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(editToolbar)
+        editToolbarConstraint = editToolbar.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: 50)
+        NSLayoutConstraint.activate([
+            editToolbarConstraint!,
+            editToolbar.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            editToolbar.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            editToolbar.heightAnchor.constraint(equalToConstant: 50)
+        ])
+
```
Mikanbu: Extra return line.

Sources/MediaViewControllers/MediaViewController.swift
```diff
@@ -102,6 +103,9 @@ class VLCMediaViewController: VLCPagingViewController<VLCLabelCell> {
 
     override func setEditing(_ editing: Bool, animated: Bool) {
         super.setEditing(editing, animated: animated)
+
+        scrollingEnabled(!editing)
```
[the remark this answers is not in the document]
carolanitz: That was supposed to disable it indeed. And for me it didn't scroll Oo. I'll double check

Sources/VLCEditController.swift
```diff
@@ -159,6 +136,10 @@ extension VLCEditController: VLCEditToolbarDelegate {
                                                                 deleteButton])
     }
 
+    func share() {
+        assertionFailure("Implement me")
```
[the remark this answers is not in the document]
carolanitz: yeah absolutely

Sources/VLCEditToolbar.swift
carolanitz: actually that filetype detection currently doesn't even work :/ I need to think of something better

Sources/VLCEditController.swift
```diff
@@ -191,15 +172,20 @@ extension VLCEditController: UICollectionViewDataSource {
...
     func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
-        if let editCell = (category as? EditableMLModel)?.editCellType() {
-            if let cell = collectionView.dequeueReusableCell(withReuseIdentifier: editCell.defaultReuseIdentifier,
-                                                             for: indexPath) as? MediaEditCell {
-                cell.media = category.anyfiles[indexPath.row]
-                cell.isChecked = selectedCellIndexPaths.contains(indexPath)
-                return cell
-            }
+        guard let editCell = (category as? EditableMLModel)?.editCellType() else {
+            assertionFailure("no editcell")
```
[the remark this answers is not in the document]
carolanitz: sure I'm going to be more detailed here. But next time add the assertions in the first place 😛 I was really confused when nothing happened when I pressed edit

Sources/VLCEditToolbar.swift
carolanitz: I looked at the code and it actually it works. I will probably have to adapt it for the allFiles tab at some point. I believe there we want to disable the button as soon as the selected cells contain one of the above mentioned file types. Which can change with every selection. So I believe I need to keep the category here and iterate over all selected files.

carolanitz: I'll make a ticket for it

Mikanbu:
Merged with 4fd8b4e836a211c561ef01a81f9cdb7ade64f150.

PR #209, opened 2019-03-07
Avoid implementing unsupported orientation method on tvOS

floatingatoll:
### Checklist

### Description
When building for tvOS, one of the iOS/tvOS shared sources attempts to implement an interface orientation method that is unavailable on tvOS. This patch skips that when the target is tvOS.

Note that I'm told the tests don't really test the tvOS side, but it builds and runs on my Apple TV before and after the patch, and without the Xcode warning.

### DIFF
--- a/Sources/VLCPlaybackNavigationController.m
+++ b/Sources/VLCPlaybackNavigationController.m
@@ -14,9 +14,11 @@
 
 @implementation VLCPlaybackNavigationController
 
+#if !TARGET_OS_TV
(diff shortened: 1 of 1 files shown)


carolanitz:
hey great spotting. I believe we can just remove the entire class from vlc-tvOS instead of adding that check :)

carolanitz:
Thanks again for pointing it out. This got addressed with 
 commit bfcd413c14c86e9bcf89c9547caa5aaae6af96da so closing.

floatingatoll:
Sure thing! Glad I could help.

PR #210, opened 2019-03-07
Import VLCTVConstants.h instead when building tvOS

floatingatoll:
### Checklist

### Description
Xcode has been objecting to redefinition of kSupportedFileFormats, which stems from VLCConstants.h being imported by VLCPlaybackController.m when it should be using VLCTVConstants.h instead.

This patch conditionally loads the correct Constants or TVConstants file, rather than assuming Constants is correct, which _seems_ to repair the error.

I am not at all certain this is the most correct approach, but in any case on the theory that fewer Xcode warnings are better over time, duly proposed.

### DIFF
--- a/Sources/VLCPlaybackController.m
+++ b/Sources/VLCPlaybackController.m
@@ -20,7 +20,11 @@
 #import "UIDevice+VLC.h"
 #import <AVFoundation/AVFoundation.h>
 #import "VLCPlayerDisplayController.h"
+#if TARGET_OS_TV
+#import "VLCTVConstants.h"
+#else
 #import "VLCConstants.h"
+#endif
 #import "VLCRemoteControlService.h"
 #import "VLCMetadata.h"
 #if TARGET_OS_IOS


carolanitz:
I need to do some testing here to make sure that nothing is needed from that file for tvOS. TVVLCConstants is already imported in the prefix header so we might want to either move VLCConstants.h over the iOS equivalent as well or remove the TVVLCConstants from the header and import it where needed

floatingatoll:
It looks like the supported extensions list has desynced between the two
(.vlc). It seemed like time drift rather than intentional removal, but I
don't have enough context to be sure.

fkuehne:
IMHO, the headers should be included by the respective prefix header for the relevant port. The desychronization of listed extensions is unintentional and should be fixed.

carolanitz:
Thanks @floatingatoll for finding another cause for a warning. Closing this now since this was addressed by f2f7743869191d54feba2a52bd89b80c173f9e5f . You're more than welcome to update the PR after feedback so that I can merge your commits instead and credit you :)

floatingatoll:
Honestly, I wasn't entirely sure how to do that yet — but I appreciate
the vote of confidence!

PR #211, opened 2019-03-08
Edit polishing

carolanitz:
### Checklist

### Description

minor updates and yes you can also add the contents of an entire playlist to an existing playlist :)

### DIFF
--- a/Resources/en.lproj/Localizable.strings
+++ b/Resources/en.lproj/Localizable.strings
@@ -56,7 +56,6 @@
 "BUTTON_ALL" = "Select All";
 "BUTTON_ABOUT" = "About";
 "BUTTON_BACK" = "Back";
-"BUTTON_EDIT" = "Edit";
 "BUTTON_DONE" = "Done";
 "BUTTON_CANCEL" = "Cancel";
 "BUTTON_SAVE" = "Save";
@@ -321,6 +320,12 @@
 //EDIT
 "ADD_TO_PLAYLIST" = "Add to Playlist";
 
+"PLAYLIST_PLACEHOLDER" = "Playlist title";
+"PLAYLIST_DESCRIPTION" = "Choose a title for your new Playlist";
+
+"RENAME_PLACEHOLDER" = "New title";
+"ERROR_RENAME_FAILED" = "Renaming failed";
+"ERROR_EMPTY_NAME" = "The title can't be empty";
 // VLCMediaLibraryKit - Sorting Criteria
 
 "ALPHA" = "Alphanumerically";
--- a/Sources/VLCEditController.swift
+++ b/Sources/VLCEditController.swift
@@ -42,13 +42,16 @@ private extension VLCEditController {
 
     private struct TextFieldAlertInfo {
         var alertTitle: String
+        var alertDescription: String
         var placeHolder: String
         var confirmActionTitle: String
 
         init(alertTitle: String = "",
+             alertDescription: String = "",
              placeHolder: String = "",
              confirmActionTitle: String = NSLocalizedString("BUTTON_DONE", comment: "")) {
             self.alertTitle = alertTitle
+            self.alertDescription = alertDescription
             self.placeHolder = placeHolder
             self.confirmActionTitle = confirmActionTitle
         }
@@ -57,7 +60,7 @@ private extension VLCEditController {
     private func presentTextFieldAlert(with info: TextFieldAlertInfo,
                                        completionHandler: @escaping (String) -> Void) {
         let alertController = UIAlertController(title: info.alertTitle,
-                                                message: "",
+                                                message: info.alertDescription,
                                                 preferredStyle: .alert)
 
         alertController.addTextField(configurationHandler: {
--- a/Sources/VLCEditController.swift
+++ b/Sources/VLCEditController.swift
@@ -150,7 +150,7 @@ extension VLCEditController: VLCEditToolbarDelegate {
             if let media = model.anyfiles[indexPath.row] as? VLCMLMedia {
                 // Not using VLCAlertViewController to have more customization in text fields
                 let alertInfo = TextFieldAlertInfo(alertTitle: String(format: NSLocalizedString("RENAME_MEDIA_TO", comment: ""), media.title),
-                                                   placeHolder: "NEW_NAME",
+                                                   placeHolder: NSLocalizedString("RENAME_PLACEHOLDER", comment: ""),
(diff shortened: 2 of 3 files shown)


SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
```diff
@@ -20,6 +20,8 @@ protocol MediaLibraryBaseModel {
     var indicatorName: String { get }
     var cellType: BaseCollectionViewCell.Type { get }
 
+    var medialibrary: VLCMediaLibraryManager { get }
+
```
Mikanbu: I fail to understand the motivation to expose the medialibrary. Could you elaborate on this?

Sources/VLCEditToolbar.swift
```diff
@@ -10,10 +10,10 @@
...
 
 protocol VLCEditToolbarDelegate: class {
-    func delete()
-    func createPlaylist()
-    func rename()
-    func share()
+    func delete(_ edittoolbar: VLCEditToolbar)
+    func addToPlaylist(_ edittoolbar: VLCEditToolbar)
+    func rename(_ edittoolbar: VLCEditToolbar)
+    func share(_ edittoolbar: VLCEditToolbar)
```
Mikanbu: I believe that we should rename the delegate methods to follow the naming convention a bit more.

See `UIApplicationDelegate`.

Resources/en.lproj/Localizable.strings
```diff
@@ -320,3 +319,10 @@
 
 //EDIT
 "ADD_TO_PLAYLIST" = "Add to Playlist";
+
+"PLAYLIST_PLACEHOLDER" = "Playlist title";
+"PLAYLIST_DESCRIPTION" = "Choose a title for your new Playlist";
+
+"RENAME_PLACEHOLDER" = "New Title";
```
Mikanbu: Not sure about the capitalization of the word playlist and title here.

Perhaps we should consider choosing either one.

SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
carolanitz: > I find a bit confusing that PR #212 and PR #213 are based on this PR.
> 
> I think PRs should be as independent as possible.

This is done on purpose so that you can check it out. You're still able to review just the last commit by selecting it in the diff

carolanitz:
> I find a bit confusing that PR #212 and PR #213 are based on this PR.
> 
> I think PRs should be as independent as possible.

This is done on purpose so that you can check it out. You're still able to review just the last commit by selecting it in the diff

Resources/en.lproj/Localizable.strings
carolanitz: you can change it if you like, I don't care I just saw that these were not localized so I fixed it.

Sources/VLCEditToolbar.swift
carolanitz: suggestion? Again here I saw that even though a delegate was declared it didn't send the calling object and bettered that

Sources/VLCEditController.swift
```diff
@@ -86,33 +89,23 @@ private extension VLCEditController {
...
+        let alertInfo = TextFieldAlertInfo(alertTitle: NSLocalizedString("PLAYLISTS", comment: ""),
+                                           alertDescription: NSLocalizedString("PLAYLIST_DESCRIPTION", comment: ""),
+                                           placeHolder: NSLocalizedString("PLAYLIST_PLACEHOLDER", comment:""))
+
+        presentTextFieldAlert(with: alertInfo, completionHandler: {
+            [selectedCellIndexPaths, model] text -> Void in
+            let playlist = model.medialibrary.createPlaylist(with: text)
+            for indexPath in selectedCellIndexPaths {
+                if let media = model.anyfiles[indexPath.row] as? VLCMLMedia {
+                    playlist.appendMedia(withIdentifier: media.identifier())
```
Mikanbu: I realized that this removes the ability to create an empty playlist.

Do we want to remove this?

SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
Mikanbu: I was thinking about this, a solution could be the following.

To avoid to expose the medialibrary, we could have a `createPlaylist` method in the models that can create a playlist.

 It could for example have a prototype like so:

`func createPlaylisty(_ name: String, filesIndexes: Set<IndexPath>? = nil)`

This would mean that the said model is responsible for the media addition in the playlist. That can help for cases such as `Genre` (it could add all media in the selected genres in a playlist).

Sources/VLCEditToolbar.swift
Mikanbu: My suggestion is still `UIApplicationDelegate` so having something like:

`applicationDidEnterBackground(_:)`

Sources/VLCEditController.swift
carolanitz: Yes for now we want to remove it BUT I have thought about this because we lack not only the ability to create an empty playlist but also the ability to add to existing ones. hence I wrote : https://code.videolan.org/videolan/vlc-ios/issues/411

carolanitz: hm thinking about it, having to present a viewcontroller makes another good case for making the Edittoolbarcontroller a UIViewcontroller and a childviewController of the MediaCategoryViewController

SharedSources/MediaLibraryModel/MediaLibraryBaseModel.swift
carolanitz: so I was going down that road for a bit until I realised that we will need to have the existing Playlists as well, to show them as selection options. A model editing another model doesn't feel right. Which means the controller needs to have access to a medialibrary if not through the model then by giving it the services.

Sources/VLCEditController.swift
Mikanbu: > we lack not only the ability to create an empty playlist ...

Just to be sure, by "we lack" do you mean UI-wise or medialibrary-wise? 🤔

Mikanbu:
Just a last nitpick, the parameter of the toolbar delegate is missing a capitalization. 😅

carolanitz:
closed by fc951ebbc8638822e1e447e3932784d3803bf0ac & b750a1d8d96d26039eca4557fa2309c5bb805d40 & dbd1ef54e83ddb999d48061df0b1416e33ad3932

PR #212, opened 2019-03-08
Add sharing

carolanitz:
### Checklist

### Description

only the add share sheet commit should be considered. 
Something that I currently very much dislike is that we present the sheet from the RootViewController but given the current architecture I can't get ahold of a UIViewController

Ticket: Sharing doesn't work yet for Album and Genre and Artist


carolanitz:
@Mikanbu I would prefer a counter solution rather than you entirely taking it from me. Maybe you can adjust it your way and send me your solution (patch) and we both talk about pros and cons?

carolanitz:
this was merged with 1d669824e92bb5950b3c438b3bb2868bf3704040

PR #213, opened 2019-03-08
Add edit cell for audio

carolanitz:
### Checklist

### Description

Currently the app crashes in some category when you hit edit to fix that I've adjusted the MediaEditCell to also handle all of our Audio models.
Once this is merged we have to write tickets for the todos in the last commit like adding file sizes and number of tracks for albums and removing the nil check for artist name. 
additionally for the Songs category the edit cell doesn't configure correctly based on the actual item type


Mikanbu:
Merged with 631bf81e.

PR #216, opened 2019-03-11
Fix migration

carolanitz:
### Checklist

### Description

follow up to https://github.com/videolan/vlc-ios/pull/190 
The tests discovered unbalanced calls in the migrationController that led to the migration controller never getting dismissed on a clean install. Two underlying issues were addressed here

1. There were two migration calls. prepareMigrationIfNeeded that showed the migration controller and startMigrationIfNeeded. This commit merges both calls to show the migration controller when the migration is actually happening.
 
2. After taking  close look at setupMedialibrary I noticed that the file discovery was being kicked off before the delegate was set since the medialibrary was lazily loaded. Therefor events were not getting received. The medialib is now immediately instantiated instead of lazily loaded.

Blocking the UI during migration is a relict from core data times where touching the database during migration crashes the system and should therefor be avoided. We have now the case where we start discovery and if we discover an old library show the migration controller with a delay after discovery is finished. 

Even though it's correct, UX wise this feels off and should be addressed in a follow up. 

Rough thoughts for follow up ticket. Two ways to address this:
1. We show the full migrationcontroller also during discovery but have the text reflect what's going on "initializing Medialib"  then "found old medialib now migrating". reason: We shouldn't show a user who made a new install a "migration view" even though there is nothing to migrate.

2. And I'd prefer this option. I believe there is no need anymore to block the UI but I would still want to give visual feedback of "initialising Medialib and migrating old data" so that when someone upgrades he understand why initially there is an empty Medialib.

Anyway there will be two Use cases that need to be covered 1. a new user 2. an old user with existing media and both deserve a great UX when first opening 3.2

### DIFF
--- a/SharedSources/VLCMediaLibraryManager.swift
+++ b/SharedSources/VLCMediaLibraryManager.swift
@@ -79,7 +79,7 @@ class VLCMediaLibraryManager: NSObject {
     private static let migrationKey: String = "MigratedToVLCMediaLibraryKit"
 
     private var didMigrate = UserDefaults.standard.bool(forKey: VLCMediaLibraryManager.migrationKey)
-
+    private var didFinishDiscovery = false
     // Using ObjectIdentifier to avoid duplication and facilitate
     // identification of observing object
     private var observers = [ObjectIdentifier: Observer]()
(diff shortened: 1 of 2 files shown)


SharedSources/VLCMediaLibraryManager.swift
```diff
@@ -79,21 +79,19 @@ class VLCMediaLibraryManager: NSObject {
...
-        return medialibrary
-    }()
+    private var medialib: VLCMediaLibrary!
 
     weak var migrationDelegate: MediaLibraryMigrationDelegate?
 
     override init() {
         super.init()
+        medialib = VLCMediaLibrary()
+        medialib.delegate = self
```
Mikanbu: Instead of implicitly unwrapping `medialib`, it could be initialized.

Additionally, I would consider putting `medialib.delegate` inside `setupMediaLibrary()`.

carolanitz:
merged with 107df7a02e4cad7444b5777726ef3b00470c8fcd & ebb6f7eac9e24ad9a541ac269ba1d2e1709184c4

PR #217, opened 2019-03-11
Add collection controller

carolanitz:
### Checklist

### Description

Is there a better approach than reusing the MediaCategoryViewController ? 
Instead of Creating 4 Controllers for an Album, a Playlist, an Artist, a Genre I've reused the CategoryViewController. Drawback is though that the required model wants one Cell for all medias which will be a problem for Playlists that can hold Movie and Audio cells.
I'm also unsure how to register for when files get added to an item of the above mentioned Collection types. probably have to dig into the media library here

### DIFF
--- a/SharedSources/MediaLibraryModel/AlbumModel.swift
+++ b/SharedSources/MediaLibraryModel/AlbumModel.swift
@@ -62,3 +62,9 @@ extension AlbumModel: MediaLibraryObserver {
         updateView?()
     }
 }
+
+extension VLCMLAlbum: MediaCollectionModel {
+    func files() -> [VLCMLMedia] {
+        return tracks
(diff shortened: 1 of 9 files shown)


carolanitz:
@Mikanbu rebased to take service renaming into account and mandatory Sortmodel. 
Items that will be addressed after this lands/ that need issues:
- add cells that work for all collection types and show position grabber in playlist
- figure out where to put sorting for genre and artist/ shouldn't be there for playlists and album 
- play all items of a collection
- edit

carolanitz:
closed by a1a4cd67c84a2622a0d49a8c642c5c378feb0028 & 6ca292501a3dd82e5299542139d6efaebad9017f

PR #219, opened 2019-03-14
Make edit controller a view controller

carolanitz:
### Checklist

### DIFF
--- a/Sources/MediaCategories/MediaCategoryViewController.swift
+++ b/Sources/MediaCategories/MediaCategoryViewController.swift
@@ -21,7 +21,7 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
     private var searchController: UISearchController?
     private let searchDataSource = VLCLibrarySearchDisplayDataSource()
     private lazy var editController: VLCEditController = {
-        let editController = VLCEditController(services:self.services, model: self.model)
+        let editController = VLCEditController(mediaLibraryManager:services.medialibraryManager, model: model)
         editController.delegate = self
         return editController
     }()
@@ -148,6 +148,7 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
         collectionView?.dataSource = editing ? editController : self
         collectionView?.delegate = editing ? editController : self
 
+        editController.resetSelections()
         displayEditToolbar()
         let layoutToBe = editing ? editCollectionViewLayout : UICollectionViewFlowLayout()
         collectionView?.setCollectionViewLayout(layoutToBe, animated: false, completion: {
--- a/Sources/MediaCategories/MediaCategoryViewController.swift
+++ b/Sources/MediaCategories/MediaCategoryViewController.swift
@@ -20,12 +20,12 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
     private var services: Services
     private var searchController: UISearchController?
     private let searchDataSource = VLCLibrarySearchDisplayDataSource()
-    private lazy var editController = VLCEditController(collectionView: self.collectionView!, model: self.model)
-    private lazy var editToolbar: VLCEditToolbar = {
-        let editToolbar = VLCEditToolbar(category: model)
-        editToolbar.delegate = editController
-        return editToolbar
+    private lazy var editController: VLCEditController = {
+        let editController = VLCEditController(model: self.model)
+        editController.delegate = self
+        return editController
     }()
+
     private var editToolbarConstraint: NSLayoutConstraint?
     private var cachedCellSize = CGSize.zero
 
@@ -97,19 +97,19 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
 
     @objc func themeDidChange() {
         collectionView?.backgroundColor = PresentationTheme.current.colors.background
-        editToolbar.backgroundColor = PresentationTheme.current.colors.background
+        editController.view.backgroundColor = PresentationTheme.current.colors.background
         setNeedsStatusBarAppearanceUpdate()
     }
 
     func setupEditToolbar() {
-        editToolbar.translatesAutoresizingMaskIntoConstraints = false
-        view.addSubview(editToolbar)
-        editToolbarConstraint = editToolbar.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: 60)
+        editController.view.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(editController.view)
+        editToolbarConstraint = editController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: 60)
         NSLayoutConstraint.activate([
             editToolbarConstraint!,
-            editToolbar.leadingAnchor.constraint(equalTo: view.leadingAnchor),
-            editToolbar.trailingAnchor.constraint(equalTo: view.trailingAnchor),
-            editToolbar.heightAnchor.constraint(equalToConstant: 50)
+            editController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            editController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            editController.view.heightAnchor.constraint(equalToConstant: 50)
         ])
     }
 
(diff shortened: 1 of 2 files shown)


Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -21,7 +21,7 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
     private var searchController: UISearchController?
     private let searchDataSource = VLCLibrarySearchDisplayDataSource()
     private lazy var editController: VLCEditController = {
-        let editController = VLCEditController(model: self.model)
+        let editController = VLCEditController(services:self.services, model: self.model)
```
Mikanbu: Do we need all the services here? 

If not we could just pass the one we need.

Additionally, extra `self.`.

Sources/VLCEditController.swift
```diff
@@ -94,29 +96,25 @@ private extension VLCEditController {
...
-                        playlist.appendMedia(withIdentifier: media.identifier())
-                    }
+        //Todo: replace with Viewcontroller that shows existing Playlists
+        let alertInfo = TextFieldAlertInfo(alertTitle: NSLocalizedString("PLAYLISTS", comment: ""),
+                                           alertDescription: NSLocalizedString("PLAYLIST_DESCRIPTION", comment: ""),
+                                           placeHolder: NSLocalizedString("PLAYLIST_PLACEHOLDER", comment:""))
+
+        presentTextFieldAlert(with: alertInfo, completionHandler: {
+            [weak self, selectedCellIndexPaths, model] text -> Void in
+            guard let playlist = self?.services.medialibraryManager.createPlaylist(with: text) else {
```
Mikanbu: If we want to `weak self`, there is no real need to capture `selectedCellIndexPaths` and `model`.

In this case, if I'm not wrong, since the lifetime of the alert is coupled with the VC, we could just capture `medialibraryManager`.

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -97,19 +97,19 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
...
+        editToolbarConstraint = editController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: 60)
         NSLayoutConstraint.activate([
             editToolbarConstraint!,
-            editToolbar.leadingAnchor.constraint(equalTo: view.leadingAnchor),
-            editToolbar.trailingAnchor.constraint(equalTo: view.trailingAnchor),
-            editToolbar.heightAnchor.constraint(equalToConstant: 50)
+            editController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            editController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            editController.view.heightAnchor.constraint(equalToConstant: 50)
         ])
```
Mikanbu: Shouldn't this setup logic be inside the new VC?

Perhaps the same for `editToolbarConstraint`.

Sources/VLCEditController.swift
```diff
@@ -82,7 +85,7 @@ private extension VLCEditController {
         alertController.addAction(cancelButton)
         alertController.addAction(confirmAction)
 
-        UIApplication.shared.keyWindow?.rootViewController?.present(alertController, animated: true, completion: nil)
+        present(alertController, animated: true, completion: nil)
```
Mikanbu: I think there are other usages of the `rootViewController` to present alerts.

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -148,7 +148,6 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
         collectionView?.dataSource = editing ? editController : self
         collectionView?.delegate = editing ? editController : self
 
-        editController.resetSelections()
```
Mikanbu: Is there a specific reason why this is removed?

It could be used directly inside the VC since it's a VC now.

Perhaps in `viewWillAppear`?

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -148,7 +148,6 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
...
         displayEditToolbar()
```
Mikanbu: I believe that the logic of this method should be placed inside the new VC.

Sources/VLCEditController.swift
```diff
@@ -9,15 +9,29 @@
...
+}
+
+class VLCEditController: UIViewController {
     private var selectedCellIndexPaths = Set<IndexPath>()
-    private let collectionView: UICollectionView
     private let model: MediaLibraryBaseModel
+    weak var delegate: VLCEditControllerDelegate?
+
+    override func loadView() {
+        super.loadView()
```
Mikanbu: I think that the super call shouldn't be called here.

As the documentation stated: 

```You can override this method in order to create your views manually. If you choose to do so, assign the root view of your view hierarchy to the view property. The views you create should be unique instances and should not be shared with any other view controller object. Your custom implementation of this method should not call super.```

Sources/MediaCategories/MediaCategoryViewController.swift
carolanitz: self is needed because of the lazy and I can just pass in the manager

Sources/MediaCategories/MediaCategoryViewController.swift
carolanitz: nope you can't do that because it needs to be relative to the view of the mediacategoryVC

Sources/VLCEditController.swift
carolanitz: true I can change that

Sources/VLCEditController.swift
carolanitz: I know but it's only there because it's in code where you don't have access to the controller that wants to present it

Sources/MediaCategories/MediaCategoryViewController.swift
carolanitz: it's removed because it's not needed. Every cell is unchecked on prepare for reuse

Sources/MediaCategories/MediaCategoryViewController.swift
carolanitz: That would mean that the editController needs to know about the MediaVCs view to perform it's display. Do you really want a subview reaching up to its superview? Shouldn't a view and it's viewcontroller rather control the position of its subviews?

Sources/MediaCategories/MediaCategoryViewController.swift
Mikanbu: In this case, it doesn't seem to reset.

demo

Sources/MediaCategories/MediaCategoryViewController.swift
Mikanbu: Well, now that EditController is a VC, I'm not sure why mediaCat should be the one taking care of updating the toolbar's constraint.

At least, if I understood correctly this would be what we have with this PR.

`MediaCatVC` --> `EditControllerVC` --> `EditToolbar`

So with this, I'm not sure that `MediaCatVC` should be the one updating/setting `EditToolbar`'s constraints.

Sources/MediaCategories/MediaCategoryViewController.swift
Mikanbu: Locally, XCode doesn't seem to complain about the `self.` changes.

Sources/MediaCategories/MediaCategoryViewController.swift
carolanitz: not quite, we have the editcontroller that is controlling the inside of the view but the edittoolbarview is a subview of the MediaCategory's view. What you propose is the case if we would push or present the EditControllerVC

Sources/MediaCategories/MediaCategoryViewController.swift
carolanitz: oh you're right so the array needs to be cleared but resetAllVisibleCell is not needed because of the prepare for reuse but this is still needed

carolanitz:
merged with b772cb0088b42e0beb583ea587329dc46b2ba4fd

PR #220, opened 2019-03-14
Sort: Use VLCActionSheet

Mikanbu:
### Checklist

### Description

Use `VLCActionSheet` to display sort alert.

While integrating VLCActionSheet, I relealized that it would be simpler to pass an array of supported sorting criterias.

Indeed, this way we won't encounter index syncronization issues and it will make the global usage easier.

A follow-up ticket to this PR would be to polish it with the new design(adding icons that we currently don't have for every criteria).

Closes [#415](https://code.videolan.org/videolan/vlc-ios/issues/415).


carolanitz:
looks good :)

Mikanbu:
Rebased.

PR #224, opened 2019-03-21
VLCEditController: Move add to playlist logic into models

Mikanbu:
### Checklist

### Description

This PR moves the add to playlist logic during the edit into the models.
This allows to have specific actions for different models.

Additionally, this integration only creates new playlist with a selection.
The ability to add to existing playlists will be added later on with the UI.

### DIFF
--- a/SharedSources/MediaLibraryModel/ArtistModel.swift
+++ b/SharedSources/MediaLibraryModel/ArtistModel.swift
@@ -76,6 +76,8 @@ extension ArtistModel {
     }
 }
 
+// MARK: - MediaLibraryObserver
+
 extension ArtistModel: MediaLibraryObserver {
     func medialibrary(_ medialibrary: MediaLibraryService, didAddArtists artists: [VLCMLArtist]) {
         artists.forEach({ append($0) })
--- a/SharedSources/MediaLibraryModel/ArtistModel.swift
+++ b/SharedSources/MediaLibraryModel/ArtistModel.swift
@@ -37,6 +37,27 @@ class ArtistModel: MLBaseModel {
     func delete(_ items: [VLCMLObject]) {
         preconditionFailure("ArtistModel: Cannot delete artist")
     }
+
+    func createPlaylist(_ name: String, _ fileIndexes: Set<IndexPath>? = nil) {
+        let playlist = medialibrary.createPlaylist(with: name)
+
+        guard let fileIndexes = fileIndexes else {
+            return
+        }
+
+        for index in fileIndexes  where index.row < files.count {
+            // Get all tracks from a VLCMLArtist
+            guard let tracks = files[index.row].tracks(with: .default, desc: false) else {
+                assertionFailure("ArtistModel: createPlaylist: Fail to retreive tracks.")
+                return
+            }
+
+            tracks.forEach() {
+                playlist.appendMedia(withIdentifier: $0.identifier())
+            }
+        }
+    }
+
 }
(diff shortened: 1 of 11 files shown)


SharedSources/MediaLibraryModel/ArtistModel.swift
```diff
@@ -37,6 +37,27 @@ class ArtistModel: MLBaseModel {
     func delete(_ items: [VLCMLObject]) {
         preconditionFailure("ArtistModel: Cannot delete artist")
     }
+
+    func createPlaylist(_ name: String, _ fileIndexes: Set<IndexPath>? = nil) {
+        let playlist = medialibrary.createPlaylist(with: name)
```
carolanitz: you might want to add a check for an empty playlist name  like ""

SharedSources/MediaLibraryModel/ArtistModel.swift
```diff
@@ -37,6 +37,27 @@ class ArtistModel: MLBaseModel {
...
+        guard let fileIndexes = fileIndexes else {
+            return
+        }
+
+        for index in fileIndexes  where index.row < files.count {
+            // Get all tracks from a VLCMLArtist
+            guard let tracks = files[index.row].tracks(with: .default, desc: false) else {
+                assertionFailure("ArtistModel: createPlaylist: Fail to retreive tracks.")
+                return
+            }
```
carolanitz: I would suggest having a tracks convenience method on Medialibrary here. I believe using the sorting method to access the tracks here is unintuitive. But mostly because every sorting call to this method triggers calls to the db which will be very slow for many tracks and it can potentially get out of sync between sorting calls if a file is added while creating the playlist.

SharedSources/MediaLibraryModel/VideoModel.swift
```diff
@@ -70,19 +71,14 @@ extension VideoModel: MediaLibraryObserver {
...
-extension VideoModel {
     func medialibrary(_ medialibrary: MediaLibraryService, thumbnailReady media: VLCMLMedia) {
-        for (index, file) in files.enumerated() {
-            if file == media {
-                files[index] = media
-                break
-            }
+        for (index, file) in files.enumerated() where file == media {
+            files[index] = media
+            break
```
carolanitz: 👍

SharedSources/MediaLibraryModel/ArtistModel.swift
Mikanbu: I would prefer merging this and adapting it once the changes are made on the medialibrary side if that is okay with you.

carolanitz: okay under the condition that you/we work next on it so that this gets addressed soon. We run into the same issue with the CollectionViewcontroller for genre, artist. We can't just use the sortcall everywhere where we use genres and tracks and have multiple cached versions of the sorting result :/ That basically screams for problems

Mikanbu: Actually, if we look into the C++ code, it is basically the same call, there is no "sort" call. Indeed, there is only one way of getting the value and depending on the parameter given it will change.

The integration of the tracks in album is actually incorrect since it only checks if it exists. So in the case where the album is updated, the tracks are not.

I asked Hugo about this and he recommends that the wrapper should directly give the value from the medialib without caching anything and if we want to optimize it, we should do it on the application side.

PR #230, opened 2019-04-01
Fix celllayout: More dynamic cellsizes

carolanitz:
We define an overall number of items based on the width.
We also have independent edgespacing and interitempadding closes #315

### Checklist

### Description

Dynamically shows between 2 and 5 movie cells, takes the safe area into consideration,
For movie cells it's 13.5 edge and between cell 7.5
For audio it's 1 to 4 cells  and edge 15 and between cell 5 

Simulator Screen Shot - iPhone X - 2019-04-01 at 16 39 06

Takes into consideration the notch: 

Simulator Screen Shot - iPhone X - 2019-04-01 at 16 39 02

Simulator Screen Shot - iPad Pro (12 9-inch) (2nd generation) - 2019-04-01 at 16 54 39
Simulator Screen Shot - iPad Pro (12 9-inch) (2nd generation) - 2019-04-01 at 16 53 48

### DIFF
--- a/Sources/MediaCategoryCells/ArtistCollectionViewCell.swift
+++ b/Sources/MediaCategoryCells/ArtistCollectionViewCell.swift
@@ -16,9 +16,6 @@ class ArtistCollectionViewCell: BaseCollectionViewCell {
 
     @IBOutlet weak var thumbnailView: UIImageView!
     @IBOutlet weak var titleLabel: UILabel!
-    override class var cellPadding: CGFloat {
-        return 5.0
-    }
 
     override var media: VLCMLObject? {
         didSet {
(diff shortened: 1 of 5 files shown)


Sources/MediaCategoryCells/ArtistCollectionViewCell.swift
```diff
@@ -49,13 +46,25 @@ class ArtistCollectionViewCell: BaseCollectionViewCell {
...
+        if width <= 414 { //max iPhone portrait width
+            numberOfCells = 1
+        } else if width <= 896 { // max landscape iPhone width includes small iPads in protrait
+            numberOfCells = 2
+        } else if width <= 1024 { // max landscape for small iPads
+            numberOfCells = 3
+        } else { //ipad Pro in Landscape
+            numberOfCells = 4
+        }
+
```
Mikanbu: I would put this logic into a helper method inside `BaseCollectionViewCell` since it looks like the values constant.

Also, we could consider having an enum for the values(414, 896, 1024). It could be a nice improvement for readability and usability(for example we could switch over it).

carolanitz:
> Overall LGTM! 🎉
> 
> Just small question, you said it closes [315](https://code.videolan.org/videolan/vlc-ios/issues/315) but if I'm not mistaken this PR addresses only the width and not the height. The ticket seems to talk about the height. 🤔
> 
> If it wasn't a mistake, then the reference inside the commit message is missing.

It fixes the spacing, padding and border part of 315 but not the dynamic height. I would make a followup for the height and I believe we could fix that up in a 3.3. 
The way it looks now is not perfect but releasable I'd say.

Sources/MediaCategoryCells/ArtistCollectionViewCell.swift
carolanitz: I was actually thinking about merging the artist and audio cell, then you avoid a lot more of duplication. numberOfCells differ between audio and video so I think BasecollectionViewCell might not work but please correct me here if I'm wrong. 
I like the enum idea a lot since these numbers are shared but I would want to look into .compact and alike. Is it okay if I make merging cells and enum a follow up and separate PR?

Mikanbu:
I totally agree that this should be merged and can be iterated on! 🎉 

I pointed out 315 just to be double sure if it was the right ticket we wanted to close. 😅

Sources/MediaCategoryCells/ArtistCollectionViewCell.swift
Mikanbu: Ideally, I would say that at least the enum should be part of this PR, but in the end, I don't mind either way. 

Your call~

Mikanbu:
Merged with f034375dc81b4fad4d924a1edd492f3a31b71a7d.

PR #232, opened 2019-04-02
[Blocked]Tests: Fix testCaptureVideoPlayback test by prepopulating the App wit…

carolanitz:
…h data

### Checklist

### Description

This PR adds a snapshot of application data to be used with UITesting.
This fixes the flaky capture Videoplayback test, that relied on networking to work which is in general not a good idea for testing. 

This also builds the base for Screenshotting when we submit the new UI and more in depth automated playback testing.

Instead of adding the app data to the repository we should add a runscript phase similar to VLCKit that downloads the needed file prior to testing


carolanitz:
runs locally but fails if run from ci due to an Xcode issue:
Similar issue for a provided location in the scheme: http://www.openradar.me/radar?id=4991106018181120

carolanitz:
we found a better solution, hence closing!

PR #236, opened 2019-04-08
Play all items of a collection

carolanitz:
### Checklist

### Description

before we just played one file, now it plays an entire collection no matter if it's a playlist or album.
I left the collectionsmodel commit in so that you can sync an album and try it out :) 
I know that this also plays all items of an artist or genre but I think it's okay. I actually even pondered if we shouldn't in general play all items in a viewcontroller.

### DIFF
--- a/Sources/MediaCategories/MediaCategoryViewController.swift
+++ b/Sources/MediaCategories/MediaCategoryViewController.swift
@@ -362,7 +362,14 @@ extension VLCMediaCategoryViewController {
 
     func play(media: VLCMLMedia) {
         VLCPlaybackController.sharedInstance().fullscreenSessionRequested = media.subtype() != .albumTrack
-        VLCPlaybackController.sharedInstance().play(media)
+        if let collectionModel = model as? CollectionModel, collectionModel.mediaCollection is VLCMLPlaylist || collectionModel.mediaCollection is VLCMLAlbum {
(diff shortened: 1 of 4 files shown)


SharedSources/MediaLibraryModel/AlbumModel.swift
```diff
@@ -69,6 +69,6 @@ extension VLCMLAlbum: MediaCollectionModel {
     }
 
     func files() -> [VLCMLMedia] {
-        return tracks
+        return self.tracks(with: .default, desc: true)
```
carolanitz: should use tracks()

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -362,7 +362,14 @@ extension VLCMediaCategoryViewController {
 
     func play(media: VLCMLMedia) {
         VLCPlaybackController.sharedInstance().fullscreenSessionRequested = media.subtype() != .albumTrack
-        VLCPlaybackController.sharedInstance().play(media)
+        if let collectionModel = model as? CollectionModel {
```
carolanitz: we shouldn't play all media of a genre or Artist I believe

carolanitz:
merged with 8520887286b522441df958a14aab9a552acbc046

PR #239, opened 2019-04-09
MediaCollectionViewCell: Adapt to configure also for movies in playlists

carolanitz:
closes #439

### Checklist

### Description

The design for a movie in a playlist still needs to be consistent and different from the one in the movietab, which is why we reuse the AudioCollectionCell here. Ideally we rename it to reflect the versatility of this class


carolanitz:
you should probably do the rename before this gets merged

carolanitz:
merged with 6c7b5b33037dc39d477a24b0902bd4640f013c32 & 872092e9fc569e3435c9317634c594e5f601cd12

PR #240, opened 2019-04-10
Downloadable content

carolanitz:
### Checklist

### Description

xml gets called from AppleTV and both html and xml are now updated. This pull request adapts the wifi download screen to the new medialibrary. This needs to be thoroughly tested! And 
 a ticket needs to be opened to add Episodes 
Screenshot 2019-04-09 at 17 01 53

### DIFF
--- a/Sources/VLCHTTPConnection.m
+++ b/Sources/VLCHTTPConnection.m
@@ -194,9 +194,9 @@
 - (HTTPDynamicFileResponse *)generateHttpResponseFrom:(NSArray *)media path:(NSString *)path
 {
     NSMutableArray *mediaInHtml = [[NSMutableArray alloc] initWithCapacity:media.count];
-    for (NSObject <VLCMLObject> *mo in media) {
-        if ([mo isKindOfClass:[VLCMLMedia class]]) {
-            VLCMLMedia *media = (VLCMLMedia *)mo;
+    for (NSObject <VLCMLObject> *mediaObject in media) {
+        if ([mediaObject isKindOfClass:[VLCMLMedia class]]) {
+            VLCMLMedia *media = (VLCMLMedia *)mediaObject;
             [mediaInHtml addObject:[NSString stringWithFormat:
                                     @"<div style=\"background-image:url('Thumbnail/%@')\"> \
                                     <a href=\"download/%@\" class=\"inner\"> \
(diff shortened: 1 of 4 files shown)


Sources/VLCHTTPConnection.m
```diff
@@ -142,31 +142,20 @@ - (BOOL)fileIsInDocumentFolder:(NSString*)filepath
...
-    if (moc) {
-        NSPersistentStoreCoordinator *psc = [moc persistentStoreCoordinator];
-        if (psc) {
-            NSManagedObject *mo = nil;
-            @try {
-                mo = [moc existingObjectWithID:[psc managedObjectIDForURIRepresentation:[NSURL URLWithString:filePath]] error:nil];
-            }@catch (NSException *exeption) {
-                return [[HTTPErrorResponse alloc] initWithErrorCode:404];
-            }
+    NSString *filePath = [[path stringByReplacingOccurrencesOfString:@"Thumbnail/" withString:@""] stringByAddingPercentEncodingWithAllowedCharacters:NSCharacterSet.URLFragmentAllowedCharacterSet];
```
Mikanbu: Nit-picking: From a small test locally, I believe it should be `/Thumbnail/`.

It is not a big deal since the extra `/` at the beginning will be ignored.

Sources/VLCHTTPConnection.m
```diff
@@ -142,31 +142,20 @@ - (BOOL)fileIsInDocumentFolder:(NSString*)filepath
...
-        }
-    }
-    return [[HTTPErrorResponse alloc] initWithErrorCode:404];
+    UIImage *thumbnail = [UIImage imageWithContentsOfFile:filePath];
+    if (!thumbnail) return [[HTTPErrorResponse alloc] initWithErrorCode:404];
+
+    NSData *theData = UIImageJPEGRepresentation(thumbnail, .9);
+
+    if (!theData) return [[HTTPErrorResponse alloc] initWithErrorCode:404];
+
```
Mikanbu: I might be wrong but I think the `if (!thumbnail)`  is not necessary since `UIImageJPEGRepresentation` will return nil if the thumbnail is nil.
Therefore, a single check for `theData` should reduce the code and be sufficient. 🎉

Sources/VLCHTTPUploaderController.h
```diff
@@ -15,11 +15,14 @@
...
 #import <UIKit/UIKit.h>
 
+@class MediaLibraryService;
+
 @interface VLCHTTPUploaderController : NSObject
 
 + (instancetype)sharedInstance;
 
 @property (nonatomic, readonly) BOOL isReachable;
+@property (nonatomic, readwrite) MediaLibraryService *medialibrary;
```
Mikanbu: Perhaps this could be a task for a follow-up ticket to have the service injected, and remove the usage of the shared instance. 👍

Sources/VLCHTTPConnection.m
carolanitz: I'll try again but I remember faintly that it didn't work when I removed the /

Sources/VLCHTTPUploaderController.h
carolanitz: yeah that really should be a follow up the shared instance angered me a lot here

carolanitz:
merged with b1d2aeba70fe6f517caebb3cd56c64a90a912e37 and following

PR #249, opened 2019-04-24
VLCTabBarCoordinator: Propagate childForStatusBarStyle

Mikanbu:
In the case of VLCMovieViewController, we need to propagate it in order
to always set the preferred status bar style to .lightContent

### Checklist

### Description

In the case of VLCMovieViewController, we need to propagate it in order to always set the preferred status bar style to `.lightContent`

### DIFF
--- a/Sources/AppearanceManager.swift
+++ b/Sources/AppearanceManager.swift
@@ -49,4 +49,8 @@ extension UINavigationController {
     override open var preferredStatusBarStyle: UIStatusBarStyle {
         return PresentationTheme.current.colors.statusBarStyle
     }
+
+    override open var childForStatusBarStyle: UIViewController? {
+        return self.topViewController
+    }
 }


Sources/VLCTabBarCoordinator.swift
```diff
@@ -65,3 +65,9 @@ extension UITabBarController {
         return PresentationTheme.current.colors.statusBarStyle
     }
 }
+
+extension UINavigationController {
```
carolanitz: Shouldn't we put it in the Appearance manager to the other extension?

carolanitz:
otherwise lgtm and can be merged but I would probably put it to the other extension

Sources/VLCTabBarCoordinator.swift
Mikanbu: I'm not sure to understand, would you prefer me to put in the Appearance manager?

carolanitz: yes we have already an extension for statusbar there as well :) 
```/extensions so that preferredStatusBarStyle is called on all childViewControllers otherwise this is not forwarded
extension UINavigationController {
    override open var preferredStatusBarStyle: UIStatusBarStyle {
        return PresentationTheme.current.colors.statusBarStyle
    }
}
```

carolanitz: and I would rather search there than in the tabbarcoordinator

Mikanbu: Ah, I get it now! Thank you for the clarification, I will make the changes.

Mikanbu:
Rebased and addressed review

PR #250, opened 2019-04-26
Translations for 3.1

mtrz:
I hope it's not to late for the last 3.1 translations update from Tx.
With this update the Thai translation gets its store description and it be activate for the release.


carolanitz:
merged after also removing serbien from the project it's 8e66989f5729d8467a4bca87dadc3dc8d520120b and commits before that

PR #251, opened 2019-04-27
Translations for master

mtrz:
It would be better not to close this PR after a successful commit but after the master will be released. Because there can be more updates before the release and additional PRs won't be needed if we will use this one till the next version is getting released.


carolanitz:
> It would be better not to close this PR after a successful commit but after the master will be released. Because there can be more updates before the release and additional PRs won't be needed if we will use this one till the next version is getting released.

For that use case we have the WIP label but additionally I'm unsure if we should keep a Pr open all the time if the intention is not to merge it or get feedback on it.

carolanitz:
oh hi  merged :)

PR #252, opened 2019-05-06
Playlist: Add reordering

carolanitz:
This adds a longpressgesture recognizer, only for playlists we enable the ability to move cells
we added  the notification for playlists modified and propogate that to the collectionmodel
Since there seems to be a bug in Medialibrary which doesn't seem to return the correct playlist we just update all playlists

### Checklist

### Description

There are two/three tickets that need to be filed once this is merged.
1. the cell flickers after reordering with the wrong thumbnail but then updates
2. if we reorder a currently playing playlist it will not update the current playlist yet
3. there seems to be an underlying bug in medialibrary for didmodifyPlaylist, it returns 3 playlists even though only one playlist changed, which makes it hard to just reload the playlist which changed, additionally our playlist is not part of that change


carolanitz:
merged with f64b2ba7e543b652c514559f823aadb7838df0ef

PR #256, opened 2019-05-07
Add Corespotlight

carolanitz:
### Checklist

### Description

this adds corespotlight to the app. 
It sets isspotlight enabled on startup 
it updates corespotlight when an item is played. 
it correctly handles a tap on a corespotlight item and starts playback of that item
it also observes the userdefault for passcode and sets is spotlightenabled to false when it is set

### DIFF
--- a/SharedSources/MediaLibraryModel/MediaModel.swift
+++ b/SharedSources/MediaLibraryModel/MediaModel.swift
@@ -63,3 +63,94 @@ extension VLCMLMedia {
         return integer == 0
     }
 }
+
+// MARK: - CoreSpotlight
+
+extension VLCMLMedia {
+    func coreSpotlightAttributeSet() -> CSSearchableItemAttributeSet {
+        let attributeSet = CSSearchableItemAttributeSet(itemContentType: "public.audiovisual-content")
+        attributeSet.title = title
+        attributeSet.metadataModificationDate = Date()
+        attributeSet.addedDate = Date()
+        attributeSet.duration = NSNumber(value: duration() / 1000)
+        attributeSet.streamable = 0
+        attributeSet.deliveryType = 0
+        attributeSet.local = 1
+        attributeSet.playCount = NSNumber(value: playCount())
+        if isThumbnailGenerated() {
+            attributeSet.thumbnailData = UIImage(contentsOfFile: thumbnail.path)?.jpegData(compressionQuality: 0.9)
+        }
+        attributeSet.codecs = codecs()
+        attributeSet.languages = languages()
+        if let audioTracks = audioTracks {
+            for track in audioTracks {
+                attributeSet.audioBitRate = NSNumber(value: track.bitrate())
+                attributeSet.audioChannelCount = NSNumber(value: track.nbChannels())
+                attributeSet.audioSampleRate = NSNumber(value: track.sampleRate())
+            }
+        }
+        if let albumTrack = albumTrack {
+            if let genre = albumTrack.genre {
+                attributeSet.genre = genre.name
+            }
+            if let artist = albumTrack.artist {
+                attributeSet.artist = artist.name
+            }
+            attributeSet.audioTrackNumber = NSNumber(value:albumTrack.trackNumber())
+            if let album = albumTrack.album {
+                attributeSet.artist = album.title
+            }
+        }
+
+        return attributeSet
+    }
+
+    func codecs() -> [String] {
+        var codecs = [String]()
+        if let videoTracks = videoTracks {
+            for track in videoTracks {
+                codecs.append(track.codec)
+            }
+        }
+        if let audioTracks = audioTracks {
+            for track in audioTracks {
+                codecs.append(track.codec)
+            }
+        }
+        if let subtitleTracks = subtitleTracks {
+            for track in subtitleTracks {
+                codecs.append(track.codec)
+            }
+        }
+        return codecs
+    }
+
+    func languages() -> [String] {
+        var languages = [String]()
+
+        if let videoTracks = videoTracks {
+            for track in videoTracks where track.language != "" {
+                languages.append(track.language)
+            }
+        }
+        if let audioTracks = audioTracks {
+            for track in audioTracks where track.language != "" {
+                languages.append(track.language)
+            }
+        }
+        if let subtitleTracks = subtitleTracks {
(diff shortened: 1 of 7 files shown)


Sources/VLCAppDelegate.m
```diff
@@ -92,10 +92,6 @@ - (void)setup
         [self->appCoordinator start];
     };
     [self validatePasscodeIfNeededWithCompletion:setupAppCoordinator];
-
-    BOOL spotlightEnabled = ![VLCKeychainCoordinator passcodeLockEnabled];
-    [[MLMediaLibrary sharedMediaLibrary] setSpotlightIndexingEnabled:spotlightEnabled];
-    [[MLMediaLibrary sharedMediaLibrary] applicationWillStart];
```
carolanitz: this moved into the medialibrary setup

Sources/VLCAppDelegate.m
```diff
@@ -154,12 +150,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
 
 - (BOOL)application:(UIApplication *)application willContinueUserActivityWithType:(NSString *)userActivityType
 {
-    if ([userActivityType isEqualToString:kVLCUserActivityLibraryMode] ||
-        [userActivityType isEqualToString:kVLCUserActivityPlaying] ||
-        [userActivityType isEqualToString:kVLCUserActivityLibrarySelection])
-        return YES;
-
-    return NO;
```
carolanitz: leftover from the watch

Sources/VLCSettingsController.m
```diff
@@ -192,9 +181,7 @@ - (void)updateForPasscode:(NSString *)passcode
     [VLCKeychainCoordinator setPasscodeWithPasscode:passcode error:&error];
     if (error == nil) {
         //Set manually the value to enable/disable the UISwitch.
-        BOOL passcodeEnabled = passcode != nil;
-        [[NSUserDefaults standardUserDefaults] setBool:passcodeEnabled forKey:kVLCSettingPasscodeOnKey];
-        [self updateUIAndCoreSpotlightForPasscodeSetting:passcode != nil];
```
carolanitz: turns out the default gets already set by the settingskit (noticed that because the observer was called twice) and the deletion is now handled in medialibrary so this method can go :)

SharedSources/MediaLibraryModel/MediaModel.swift
```diff
@@ -73,3 +73,75 @@ extension VLCMLMedia {
...
+        attributeSet.duration = NSNumber(value: duration() / 1000)
+        attributeSet.streamable = 0
+        attributeSet.deliveryType = 0
+        attributeSet.local = 1
+        attributeSet.playCount = NSNumber(value: playCount())
+        if isThumbnailGenerated() {
+            attributeSet.thumbnailData = UIImage(contentsOfFile: thumbnail.path)?.jpegData(compressionQuality: 0.9)
+        }
+        attributeSet.codecs = codecs()
+        attributeSet.languages = langugaes()
```
ePirat: Typo: `langugaes()` should probably be `languages()`

SharedSources/MediaLibraryModel/MediaModel.swift
```diff
@@ -73,3 +73,75 @@ extension VLCMLMedia {
...
+        for track in audioTracks where track.codec != nil {
+            codecs.append(track.codec)
+        }
+//        for track in subtitleTracks where track.codec != nil {
+//            codecs.append(track.codec)
+//        }
+        return codecs
+    }
+
+    func langugaes() -> [String] {
```
ePirat: Typo (see above)

SharedSources/MediaLibraryModel/MediaModel.swift
```diff
@@ -73,3 +73,75 @@ extension VLCMLMedia {
...
+        }
+        attributeSet.codecs = codecs()
+        attributeSet.languages = langugaes()
+        for track in audioTracks {
+            attributeSet.audioBitRate = NSNumber(value: track.bitrate())
+            attributeSet.audioChannelCount = NSNumber(value: track.nbChannels())
+            attributeSet.audioSampleRate = NSNumber(value: track.sampleRate())
+        }
+        if let albumTrack = albumTrack {
+           // attributeSet.genre = albumTrack.genre.name
```
ePirat: Why is this commented out?

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -232,11 +232,25 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
...
     }
+
+    func createSpotlightItem(media: VLCMLMedia) {
+        if KeychainCoordinator.passcodeLockEnabled {
+            return
+        }
+        userActivity = NSUserActivity(activityType: kVLCUserActivityPlaying)
+        userActivity?.title = media.title
+        userActivity?.contentAttributeSet = media.coreSpotlightAttributeSet()
+        userActivity?.userInfo = ["playingmedia" : media.identifier()]
```
ePirat: Would probably be a bit nicer to not have this magic string `playingmedia` in two places, but instead a define, like for `kVLCUserActivityPlaying`?

SharedSources/MediaLibraryModel/MediaModel.swift
```diff
@@ -73,3 +73,75 @@ extension VLCMLMedia {
...
+    func codecs() -> [String] {
+        var codecs = [String]()
+
+        for track in videoTracks where track.codec != nil {
+             codecs.append(track.codec)
+        }
+        for track in audioTracks where track.codec != nil {
+            codecs.append(track.codec)
+        }
+//        for track in subtitleTracks where track.codec != nil {
```
ePirat: Why is this commented out?

SharedSources/MediaLibraryModel/MediaModel.swift
carolanitz: because it crashes without the nullability

SharedSources/MediaLibraryModel/MediaModel.swift
carolanitz: because the subtitle tracks are going to be exposed with the new medialibrary update

PR #258, opened 2019-05-07
Skip movieDBSessionManager init when no API key is set

floatingatoll:
### Checklist

### Description
I noticed while building VLC over the past few months that browsing a network share causes a lot of Xcode console spam and 401 Unauthorized responses from themoviedb.org. I eventually tracked this down to two instances of movieDBSessionManager being configured when the API key is `""`, which results in 100% bad queries to the remote website. This commit only sets up movieDBSessionManager when the API key is not `""` so that folks who are building and running VLC without one stop generating `GET configuration?apikey=` network requests to themoviedb's site.

I don't have a valid API key, so I can't easily test this, but at the very least my local build continues to work correctly in the network file browser without generating either the 401 Unauth http request or any other new warnings/errors on the console.


Mikanbu:
Merged with a12cb2ee9631c72877cca838fb84eb7f696e97d0 and e9413d6e on 3.0.

Thanks again.

floatingatoll:
Sure thing :)

PR #261, opened 2019-05-10
Implement search

carolanitz:
### Checklist

### Description

okay this is out of wipety wip wip and the only remaining issue is getting out of search mode and well it jumps a bit when you pull down but that's okay for now


Mikanbu:
Awesome PR!

I haven't look at the code yet but it already seems to be working quite nicely!
Just a few things I noticed for now while testing a bit:

- I believe it could be nice to have a placeholder in the text field.
- Even though we don't take into account the capitalization, perhaps we should change the `UITextAutocapitalizationType`.
- It looks like it has a bit of trouble resigning first responder even on empty text field or tap away. This leads to `Sort` and `Edit` being sometimes hidden when it shouldn't.

Sources/LibrarySearchDataSource.swift
```diff
@@ -0,0 +1,105 @@
...
+            } else if let playlist = $0 as? VLCMLPlaylist {
+                if playlist.contains(lowercaseSearchString) { searchData.append($0) }
+            } else if let artist = $0 as? VLCMLArtist {
+                if artist.contains(lowercaseSearchString) { searchData.append($0) }
+            } else if let albumtrack = $0 as? VLCMLAlbumTrack {
+                if albumtrack.contains(lowercaseSearchString) { searchData.append($0) }
+            } else {
+                assertionFailure("unhandled type")
+            }
+        }
```
Mikanbu: In order to avoid this, since it seems that for now, we do the same thing for all the different types, what would you think of something like this:

Indeed, we could use a protocol here, like so:
```
protocol MLSearchable {
    func contains(_ searchString: String) -> Bool
}
```

Conforming the different VLCML* objects making this part quite simpler:

```
model.anyfiles.forEach {
    if let searchableFile = $0 as? MLSearchable,
        searchableFile.contains(lowercaseSearchString) {
        searchData.append($0)
     } else {
        assertionFailure("LibrarySearchDataSource: Unhandled type")
     }
}
```

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -17,13 +17,14 @@ protocol MediaCategoryViewControllerDelegate: NSObjectProtocol {
     func needsToUpdateNavigationbarIfNeeded(_ viewController: VLCMediaCategoryViewController)
 }
 
-class VLCMediaCategoryViewController: UICollectionViewController, UICollectionViewDelegateFlowLayout, UISearchResultsUpdating, UISearchControllerDelegate, IndicatorInfoProvider {
+class VLCMediaCategoryViewController: UICollectionViewController, UICollectionViewDelegateFlowLayout, UISearchBarDelegate, IndicatorInfoProvider {
 
     var model: MediaLibraryBaseModel
-
+    var searchBar = UISearchBar(frame: .zero)
+    private var searchbarConstraint: NSLayoutConstraint?
```
Mikanbu: What would you think of using a closure to initialize this lazily?

Indeed, this can be confusing if we don't need to actually use the optionality.

On the nit-picking side, I realize that it isn't your code but could we put the search properties grouped together?
Additionally, capitalization missing on `searchBarConstraint`.

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -81,14 +82,30 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
...
+        setupHeader()
         if let collection = model as? CollectionModel {
             title = collection.mediaCollection.title()
         }
         NotificationCenter.default.addObserver(self, selector: #selector(themeDidChange), name: .VLCThemeDidChangeNotification, object: nil)
         navigationItem.rightBarButtonItems = [editButtonItem, UIBarButtonItem(customView: rendererButton)]
     }
 
+    func setupHeader() {
+
```
Mikanbu: Should this method be renamed to `setupSearchBar`?

Sources/MediaViewControllers/MediaViewController.swift
```diff
@@ -16,6 +16,7 @@ class VLCMediaViewController: VLCPagingViewController<VLCLabelCell>, MediaCatego
 
     var services: Services
     private var rendererButton: UIButton
+
```
Mikanbu: Nit-pick: Extra unrelated changes.

Sources/PagerStripViewController.swift
```diff
@@ -96,6 +96,7 @@ class PagerTabStripViewController: UIViewController, UIScrollViewDelegate {
         }
         isViewAppearing = false
         children.forEach { $0.endAppearanceTransition() }
+        //updateSearchbarController()
```
Mikanbu: Should this be commented out or removed?

Sources/PagerStripViewController.swift
```diff
@@ -122,7 +123,6 @@ class PagerTabStripViewController: UIViewController, UIScrollViewDelegate {
             preCurrentIndex = index
             return
         }
-
```
Mikanbu: Nit-pick: Extra unrelated changes.

Sources/LibrarySearchDataSource.swift
carolanitz: yeah sure I don't care

Sources/MediaCategories/MediaCategoryViewController.swift
carolanitz: I believe it is more confusing to make it lazily intialized because you rip it out of context with the other searchbar constraints. 
I can group the properties together yes

Sources/MediaCategories/MediaCategoryViewController.swift
carolanitz: will do

Sources/LibrarySearchDataSource.swift
carolanitz: yeah I believe it's actually better indeed

PR #276, opened 2019-05-20
Collectionmodels: Delete object in file when deletion was observed

carolanitz:
### Checklist

### Description

This fixes that artists, genre and albums were not deleted. We now added observers for deletion


Mikanbu:
Is there a missing commit about `modelChanged` ? 🤔

carolanitz:
@Mikanbu ah yeah I need to change that line to updateView?() but I have one commit where I changed updateview to modelchanged with files which can then in the mediacategory be used to diff between the current and new files to animated remove / add changed files but that's for another time

PR #277, opened 2019-05-21
Sort: Hightlight current selection

Mikanbu:
### Checklist

### Description

The main purpose was to have the current sort criteria highlighted in the ActionSheet.
This includes refactoring of VLCSettinsSheetCell into ActionSheetCell so that by default we provide a highlight option.
Additionally, I have noticed that we were asking the wrong sorting criteria by default to the medialibrary which wasn't in harmony with the sorting criterias we provided in the UI.


carolanitz:
LGTM but in this PR there is _a lot_ more going on than just highlight current sort selection :D

PR #285, opened 2019-05-21
CollectionModel: Add removal from Playlist

carolanitz:
(fixes #507)

### Checklist

### DIFF
--- a/SharedSources/MediaLibraryModel/CollectionModel.swift
+++ b/SharedSources/MediaLibraryModel/CollectionModel.swift
@@ -40,7 +40,13 @@ class CollectionModel: MLBaseModel {
     var indicatorName: String = "Collections"
 
     func delete(_ items: [VLCMLObject]) {
-       assertionFailure("still needs implementation")
+        if let playlist = mediaCollection as? VLCMLPlaylist {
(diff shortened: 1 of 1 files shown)


SharedSources/MediaLibraryModel/CollectionModel.swift
```diff
@@ -40,7 +40,13 @@ class CollectionModel: MLBaseModel {
     var indicatorName: String = "Collections"
 
     func delete(_ items: [VLCMLObject]) {
-       assertionFailure("still needs implementation")
+        if let playlist = mediaCollection as? VLCMLPlaylist {
+            for item in items where item is VLCMLMedia {
+                if let index = files.firstIndex(of: item as! VLCMLMedia) {
```
Mikanbu: What do you think of something like this:

``` 
for case let media as VLCMLMedia in items {
    if let index = files.firstIndex(of: media) {
        ...
    }
```

carolanitz: didn't know that was possible :D will do

carolanitz:
merged with 2f2cfae524926e7c9abfaac6235ad04232426158

PR #290, opened 2019-05-23
VLCMediaCategoryViewController: Handle AutomaticallyPlayNextItem setting

Mikanbu:
### Checklist

### Description

Related issue [492](https://code.videolan.org/videolan/vlc-ios/issues/492).

Assertions could be considered when the files are nil.

### DIFF
--- a/Sources/MediaCategories/MediaCategoryViewController.swift
+++ b/Sources/MediaCategories/MediaCategoryViewController.swift
@@ -255,7 +255,7 @@ class VLCMediaCategoryViewController: UICollectionViewController, UICollectionVi
     // MARK: - UICollectionViewDelegate
     override func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
         if let media = model.anyfiles[indexPath.row] as? VLCMLMedia {
-            play(media: media)
+            play(media: media, at: indexPath)
             createSpotlightItem(media: media)
         } else if let mediaCollection = model.anyfiles[indexPath.row] as? MediaCollectionModel {
             let collectionViewController = VLCCollectionCategoryViewController(services, mediaCollection: mediaCollection)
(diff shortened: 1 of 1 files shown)


Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -443,16 +443,23 @@ private extension VLCMediaCategoryViewController {
 
 extension VLCMediaCategoryViewController {
 
-    func play(media: VLCMLMedia) {
-        VLCPlaybackController.sharedInstance().fullscreenSessionRequested = media.subtype() != .albumTrack
```
carolanitz: this here shouldn't be removed ^

Sources/MediaCategories/MediaCategoryViewController.swift
```diff
@@ -443,16 +443,23 @@ private extension VLCMediaCategoryViewController {
...
-                return
-            }
-            VLCPlaybackController.sharedInstance().playMedia(at: index, fromCollection: collectionModel.files)
+    func play(media: VLCMLMedia, at indexPath: IndexPath) {
+        let playbackController = VLCPlaybackController.sharedInstance()
+        let autoPlayNextItem = UserDefaults.standard.bool(forKey: kVLCAutomaticallyPlayNextItem)
+
+        guard autoPlayNextItem else {
+            playbackController.play(media)
+            return
```
carolanitz: I find this use of guard a little bit unsettling here. Guard is a keyword that I link to faulty conditions that we prepare against like ensuring something is not nil or that text is not just "" that we want to return from.
I'd be happier if it's just an if. 
It also took me quite a while to understand the code because of it

Mikanbu: I will make the change!

Sources/MediaCategories/MediaCategoryViewController.swift
Mikanbu: You're right, but I will change the detection to `media.type() != .audio` which makes it easier to read and understand.

I was a bit confused of why it was only albumTrack at first.

Mikanbu:
Addressed the review and rebased to master.

Sources/MediaCategories/MediaCategoryViewController.swift
carolanitz: test it though because I think it wasn't always working

PR #293, opened 2019-05-23
MediaCollectionViewCell: use bigger thumbnails on ipad

carolanitz:
### Checklist

### DIFF
--- a/Sources/MediaCategoryCells/MediaCollectionViewCell.swift
+++ b/Sources/MediaCategoryCells/MediaCollectionViewCell.swift
@@ -14,10 +14,11 @@ import Foundation
 
 class MediaCollectionViewCell: BaseCollectionViewCell {
 
-    @IBOutlet weak var thumbnailView: UIImageView!
-    @IBOutlet weak var titleLabel: UILabel!
-    @IBOutlet weak var descriptionLabel: UILabel!
-    @IBOutlet weak var newLabel: UILabel!
+    @IBOutlet private weak var thumbnailView: UIImageView!
+    @IBOutlet private weak var titleLabel: UILabel!
+    @IBOutlet private weak var descriptionLabel: UILabel!
+    @IBOutlet private weak var newLabel: UILabel!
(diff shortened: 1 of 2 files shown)


Sources/MediaCategoryCells/MediaCollectionViewCell.swift
```diff
@@ -18,6 +18,7 @@ class MediaCollectionViewCell: BaseCollectionViewCell {
     @IBOutlet weak var titleLabel: UILabel!
     @IBOutlet weak var descriptionLabel: UILabel!
     @IBOutlet weak var newLabel: UILabel!
+    @IBOutlet weak var thumbnailWidth: NSLayoutConstraint!
```
Mikanbu: Ideally, I believe the IBOutlet should be set as private.

Sources/MediaCategoryCells/MediaCollectionViewCell.swift
```diff
@@ -43,6 +44,8 @@ class MediaCollectionViewCell: BaseCollectionViewCell {
         super.awakeFromNib()
         newLabel.text = NSLocalizedString("NEW", comment: "")
         newLabel.textColor = PresentationTheme.current.colors.orangeUI
+         let isIpad = UIDevice.current.userInterfaceIdiom == UIUserInterfaceIdiom.pad
```
Mikanbu: Nit-pick: Extra space

Mikanbu:
I was also wondering, shouldn't the images, colour(theme) be done before showing it to the users? 🤔 

For example, if we switch to the dark theme, we can see the default images change.

Sources/MediaCategoryCells/MediaCollectionViewCell.swift
carolanitz: can do and will do for all then

carolanitz:
rebased and adjusted. the last one regarding you can see the default image change wasn't fixed. i can create a follow up ticket for that

carolanitz:
merged with 35eebb8ad19a126269e26bc6101090c82f5b2cff

PR #295, opened 2019-05-24
MediaCollectionViewCell: Handle unknown and various artist

Mikanbu:
### Checklist

### Description

In the medialibrary, there is a specific ID for unknown and various artist.

This uses them in order to show the correct localization for them.

### DIFF
--- a/Sources/MediaCategoryCells/MediaCollectionViewCell.swift
+++ b/Sources/MediaCategoryCells/MediaCollectionViewCell.swift
@@ -60,21 +60,21 @@ class MediaCollectionViewCell: BaseCollectionViewCell {
         thumbnailView.layer.masksToBounds = true
         thumbnailView.layer.cornerRadius = thumbnailView.frame.size.width / 2.0
         titleLabel.text = audiotrack.title
-        descriptionLabel.text = audiotrack.albumTrack?.artist?.name ?? NSLocalizedString("UNKNOWN_ARTIST", comment: "")
+        descriptionLabel.text = audiotrack.albumTrackArtistName()
         newLabel.isHidden = !audiotrack.isNew
         thumbnailView.image = audiotrack.thumbnailImage()
     }
 
     func update(album: VLCMLAlbum) {
-        titleLabel.text = album.title != "" ? album.title : NSLocalizedString("UNKNOWN_TITLE", comment: "")
-        descriptionLabel.text = album.albumArtist?.name != "" ? album.albumArtist?.name : NSLocalizedString("UNKNOWN_ARTIST", comment: "")
+        titleLabel.text = album.albumName()
+        descriptionLabel.text = album.albumArtistName()
         thumbnailView.image = album.thumbnail()
     }
 
     func update(artist: VLCMLArtist) {
         thumbnailView.layer.masksToBounds = true
         thumbnailView.layer.cornerRadius = thumbnailView.frame.size.width / 2.0
-        titleLabel.text = artist.name
+        titleLabel.text = artist.artistName()
         descriptionLabel.text = artist.numberOfTracksString()
         thumbnailView.image = artist.thumbnail()
     }
--- a/Podfile
+++ b/Podfile
@@ -23,7 +23,7 @@ target 'VLC-iOS' do
   pod 'PAPasscode', '~>1.0'
   pod 'GoogleAPIClientForREST/Drive'
   pod 'MobileVLCKit', '3.3.2'
-  pod 'VLCMediaLibraryKit', '0.2.1'
+  pod 'VLCMediaLibraryKit', '0.2.2'
   pod 'MediaLibraryKit-prod'
   pod 'GTMAppAuth'
   pod 'OneDriveSDK'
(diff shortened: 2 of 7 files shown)


Sources/MediaCategoryCells/MediaCollectionViewCell.swift
```diff
@@ -132,3 +132,21 @@ class MediaCollectionViewCell: BaseCollectionViewCell {
...
+// MARK: - Private helpers
+
+private extension MediaCollectionViewCell {
+    private func name(of artist: VLCMLArtist?) -> String {
+        guard let artist = artist else {
+            return NSLocalizedString("UNKNOWN_ARTIST", comment: "")
+        }
+
+        if artist.identifier() == UnknownArtistID {
+            return NSLocalizedString("UNKNOWN_ARTIST", comment: "")
```
Mikanbu: I realize that this might be considered as duplication and can be put with the guard. I decided mostly for readability to have a separate condition.

Sources/MediaCategoryCells/MediaCollectionViewCell.swift
```diff
@@ -57,21 +57,21 @@ class MediaCollectionViewCell: BaseCollectionViewCell {
...
-        descriptionLabel.text = audiotrack.albumTrack?.artist?.name ?? NSLocalizedString("UNKNOWN_ARTIST", comment: "")
+        descriptionLabel.text = name(of: audiotrack.albumTrack?.artist)
         newLabel.isHidden = !audiotrack.isNew
         thumbnailView.image = audiotrack.thumbnailImage()
     }
 
     func update(album: VLCMLAlbum) {
-        titleLabel.text = album.title != "" ? album.title : NSLocalizedString("UNKNOWN_TITLE", comment: "")
-        descriptionLabel.text = album.albumArtist?.name != "" ? album.albumArtist?.name : NSLocalizedString("UNKNOWN_ARTIST", comment: "")
+        titleLabel.text = album.isUnknownAlbum() ? NSLocalizedString("UNKNOWN_ALBUM", comment: "") : album.title
```
carolanitz: I would hide that behind a view model function/ computed property that we use everywhere  instead of having these checks everywhere

Sources/MediaCategoryCells/MediaCollectionViewCell.swift
```diff
@@ -57,21 +57,21 @@ class MediaCollectionViewCell: BaseCollectionViewCell {
         thumbnailView.layer.masksToBounds = true
         thumbnailView.layer.cornerRadius = thumbnailView.frame.size.width / 2.0
         titleLabel.text = audiotrack.title
-        descriptionLabel.text = audiotrack.albumTrack?.artist?.name ?? NSLocalizedString("UNKNOWN_ARTIST", comment: "")
+        descriptionLabel.text = name(of: audiotrack.albumTrack?.artist)
```
carolanitz: the function itself is already a good idea. I would just have rather a view model property on audiotrack

Sources/MediaCategoryCells/MediaCollectionViewCell.swift
carolanitz: yeah that's fine :)

Sources/MediaCategoryCells/MediaCollectionViewCell.swift
Mikanbu: I agree that it should be moved to somewhere, I'm not sure if it should inside a view model.

What would you think of putting it as a helper method of, for example, `BaseCollectionViewCell`?

This would have the benefit of keeping the same prototype accepting a `VLCMLArtist?` which removes a check from at the usage for example by using `name(of: audiotrack.albumTrack?.artist)` you won't need to check for nullability.

If not, we would put it as an extenstion of `VLCMLArtist` but that would mean we need to check for nullability each time we need to show the name of an artist from "something else" than a `VLCMLArtist`.
e.g: `descriptionLabel.text = album.albumArtist?.localizedName() ?? NSLocalizedString("UNKNOWN_ARTIST", comment: "")`

Or we could have both ways, meaning having a helper inside `BaseCollectionViewCell` which checks for nullability with the usage of an extension in `VLCMLArtist`?

I do not mind either ways, I might prefer option 3 for usability.

carolanitz: There should be nothing like ```album.albumArtist?.localizedName() ?? NSLocalizedString("UNKNOWN_ARTIST", comment: "")``` in the views or viewcontroller

It belongs in the model and to the view model methods because that's where we handle how certain states of the model should be displayed everywhere. If you decide one day to display that object in a popup, header or where ever, you don't want to have to remember and duplicate the code above with the check.

Therefore I suggest going with the computed property or function (maybe called artistName instead of localisedName since your not translating an artists name :)) on VLCMLArtist and return from there either name or localised "unknown artist". On VLCMLAlbum same thing but maybe it's called albumArtistame. If album has an artist return artist.artistName or return a localised version for not having an artist on an album.

In our case it is "Unknown Artist" in both cases but there are other instances where you might want to distinguish later and say "Unknown Album Artist" vs "Unknown Artist".
While it is a bit more code on the model side it keeps or view controllers clean.

(Here is a good read about MVVM, though we're not doing it exactly like that but it explains the rational behind it: https://blog.pusher.com/mvvm-ios/)

Mikanbu:
Fixup & Rebased to master

Mikanbu:
Small mistake, rebased to last commits on master.

PR #298, opened 2019-05-26
Do not invert colors of video and images

dcordero:
### Checklist

### Description

Fix [#501](https://code.videolan.org/videolan/vlc-ios/issues/501)

### DIFF
--- a/Sources/MiniPlayer/AudioMiniPlayer.swift
+++ b/Sources/MiniPlayer/AudioMiniPlayer.swift
@@ -61,7 +61,7 @@ private extension AudioMiniPlayer {
 
         artworkImageView.clipsToBounds = true
         artworkImageView.layer.cornerRadius = 2
-
+        
         playPauseButton.accessibilityLabel = NSLocalizedString("PLAY_PAUSE_BUTTON", comment: "")
         nextButton.accessibilityLabel = NSLocalizedString("NEXT_BUTTON", comment: "")
         previousButton.accessibilityLabel = NSLocalizedString("PREV_BUTTON", comment: "")
@@ -180,6 +180,10 @@ private extension AudioMiniPlayer {
         artistLabel.text = metadata.artist
         if metadata.isAudioOnly {
             artworkImageView.image = metadata.artworkImage ?? UIImage(named: "no-artwork")
+            
+            if #available(iOS 11.0, *) {
+                artworkImageView.accessibilityIgnoresInvertColors = metadata.artworkImage != nil
+            }
         }
     }
 }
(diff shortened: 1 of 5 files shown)


Sources/MiniPlayer/AudioMiniPlayer.swift
```diff
@@ -61,7 +61,7 @@ private extension AudioMiniPlayer {
 
         artworkImageView.clipsToBounds = true
         artworkImageView.layer.cornerRadius = 2
-
+
```
Mikanbu: Nit-pick: Extra unnecessary changes.

Sources/MiniPlayer/AudioMiniPlayer.swift
```diff
@@ -180,6 +180,10 @@ private extension AudioMiniPlayer {
         artistLabel.text = metadata.artist
         if metadata.isAudioOnly {
             artworkImageView.image = metadata.artworkImage ?? UIImage(named: "no-artwork")
+            
+            if #available(iOS 11.0, *) {
+                artworkImageView.accessibilityIgnoresInvertColors = metadata.artworkImage != nil
```
Mikanbu: Shouldn't the `artworkImageView` always be set to true since in the case where `metadata.artworkImage` is nil, we are setting `artworkImageView.image` to a default artwork?

Sources/MiniPlayer/AudioMiniPlayer.swift
dcordero:
```suggestion

```

Sources/MiniPlayer/AudioMiniPlayer.swift
dcordero: I do not think so. The asset [no-artwork](https://github.com/videolan/vlc-ios/blob/d338742a878a6dda59e2fd7781097836802d955e/vlc-ios/Images.xcassets/MediaController/no-artwork.imageset/no-artwork%402x.png) is a plain image, inverting the colors should not be destructive on it.

Mikanbu: Indeed, I agree with you for the no-artwork!

Sadly I just thought of a special case. For the first version of the new UI we decided to use the AudioMiniPlayer as an VideoMiniPlayer as well! At least for now, this would means that we probably always need to set `accessibilityIgnoresInvertColors`.

Mikanbu:
Rebased and merged with 740285e409f4cbd715bd075a7356d7538bc82af1.

I took the liberty to add the related ticket into the commit message.

Thank you for your contribution! 🎉

PR #300, opened 2019-05-29
HTTP Import: Respect folder structure

Mikanbu:
### Checklist

### Description

This might not be the ideal solution since I have a very limited knowledge on our HTTP-fun(JS, JQuery, ...) side.
With this patch, we now send the whole path in the `MultipartMessageHeader` which makes thinks easier on the application side to handle the file structure.

Testing are welcome!

Closes [#381](https://code.videolan.org/videolan/vlc-ios/issues/381) and partially avoids [medialibrary#74](https://code.videolan.org/videolan/medialibrary/issues/74).

### DIFF
--- a/Sources/VLCHTTPUploaderController.m
+++ b/Sources/VLCHTTPUploaderController.m
@@ -303,22 +303,40 @@
      * while on iOS we have persistent storage, so move it there */
 #if TARGET_OS_IOS
     NSString *fileName = [filepath lastPathComponent];
-    NSArray *searchPaths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
-    NSString *libraryPath = searchPaths[0];
-    NSString *finalFilePath = [libraryPath stringByAppendingPathComponent:fileName];
+    NSString *libraryPath = [NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES) firstObject];
+    NSString *uploadPath = [[NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES)
+                            firstObject] stringByAppendingPathComponent:@"Upload"];
+
+    NSString *finalFilePath = [libraryPath
+                               stringByAppendingString:[filepath
+                                                        stringByReplacingOccurrencesOfString:uploadPath
+                                                        withString:@""]];
+
     NSFileManager *fileManager = [NSFileManager defaultManager];
+
+    // Re-create the folder structure of the user
+    if (![fileManager createDirectoryAtPath:[finalFilePath stringByDeletingLastPathComponent]
+                withIntermediateDirectories:YES attributes:nil error:nil])
+        APLog(@"Could not create directory at path: %@", finalFilePath);
 
     if ([fileManager fileExistsAtPath:finalFilePath]) {
         /* we don't want to over-write existing files, so add an integer to the file name */
-        NSString *potentialFilename;
+        NSString *potentialFullPath;
+        NSString *currentPath = [finalFilePath stringByDeletingLastPathComponent];
         NSString *fileExtension = [fileName pathExtension];
         NSString *rawFileName = [fileName stringByDeletingPathExtension];
         for (NSUInteger x = 1; x < 100; x++) {
-            potentialFilename = [NSString stringWithFormat:@"%@ %lu.%@", rawFileName, (unsigned long)x, fileExtension];
-            if (![[NSFileManager defaultManager] fileExistsAtPath:[libraryPath stringByAppendingPathComponent:potentialFilename]])
+            potentialFullPath = [currentPath stringByAppendingString:[NSString
+                                                                      stringWithFormat:@"/%@-%lu.%@",
+                                                                      rawFileName,
+                                                                      (unsigned long)x,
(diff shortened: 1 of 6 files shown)


Sources/VLCHTTPUploaderController.m
```diff
@@ -303,22 +303,40 @@ - (void)moveFileFrom:(NSString *)filepath
      * while on iOS we have persistent storage, so move it there */
 #if TARGET_OS_IOS
     NSString *fileName = [filepath lastPathComponent];
-    NSArray *searchPaths = NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES);
-    NSString *libraryPath = searchPaths[0];
-    NSString *finalFilePath = [libraryPath stringByAppendingPathComponent:fileName];
+    NSString *libraryPath = [NSSearchPathForDirectoriesInDomains(NSDocumentDirectory, NSUserDomainMask, YES) firstObject];
+    NSString *uploadPath = [[NSSearchPathForDirectoriesInDomains(NSCachesDirectory, NSUserDomainMask, YES)
+                            firstObject] stringByAppendingPathComponent:@"Upload"];
```
carolanitz: can we have a constant for "Upload" because I believe it's used somewhere else as well

Sources/VLCHTTPUploaderController.m
```diff
@@ -303,22 +303,40 @@ - (void)moveFileFrom:(NSString *)filepath
...
+
+    NSString *finalFilePath = [libraryPath
+                               stringByAppendingString:[filepath
+                                                        stringByReplacingOccurrencesOfString:uploadPath
+                                                        withString:@""]];
```
carolanitz: can we use a better/ more expressive name than final? Something that describes what we're doing here? If I see it correctly you're getting the relative filepath without the created uploadpath by us correct?

Mikanbu:
> What would happen with an empty folder? 

This behaviour doesn't change with this change.
Unless there is at least one file in the hierarchy, the structure is not imported.

For example: 

If the user imports:

```
      .
      ├── a
      │   ├── b
      │   │   └── c
```

Nothing will be imported.

In this following case:

```
      .
      ├── a
      │    ├── b
      │    │   └── c
      |    |       └── bongo_cat.mp4
```

The whole folder structure is imported with the mp4 file.

There are edge cases that I didn't dive too deep into such as:

```
      .
      ├── a
      │    ├── b
      │    │   └── c
      |    |       └── bongo_cat.mp4
      |    |   └── d
```

In the case above, the whole structure is imported even the folder `d`.
We could add a check if the folder has content, perhaps in a follow-up.
Additionally, importing duplicate folders isn't gracefully handled yet.
If the user imports the same folder twice, the content of the folders will be put in the same place with different names. Ideally, it should be a separate folder.

> not only synced over but after deletion of the last file in it?

If the user imported a folder with a single file and the said file is deleted, the folder stays.

>  Does it stay around forever or is there a way to delete it?

Yes and yes, with, for example, Files.

> Also a user expects that a playlist is created if a folder is synced at least looking at the ticket it seems so. That would open pandoras box with nested folders and open the question if playlist creation should lead to a folder creation on disk? but since deleting a playlist doesn't delete it's files it might not be the right equivalent?

We do not create a playlist when a folder is imported. At least not in this PR.
This PR is a solution to [medialibrary#74](https://code.videolan.org/videolan/medialibrary/issues/74).

Please understand that there is a difference between *playlist* and *grouping of media*.

Indeed, correct me if I'm wrong but it seems that you are referring more to a grouping of media more than a playlist itself, and that is only for videos.

For example, on a medialibrary level, if a user imports albums with a folder structure:

```
      ├── Album_01
      │   │   └── Tracks...
      ├── Album_02
      │   │   └── Tracks...
```

It wouldn't make much sense to create a playlist for each album, since the medialibrary already expose them separately.

For videos on the other hand, the grouping of media is something that we need to add later on, and this has not much to do with this PR.
The idea will be that, on the application side, we need to handle `IFolder` objects so that we can expose to users a grouping of media.


> ... look how android handles it ...

Android doesn't have HTTP WiFi sharing. Additionally, it is more complex on Android since there is external storage on different devices. Therefore having a playlist created automatically on every folder doesn't seem to be ideal in my opinion.
If I'm not mistaken, there is a distinction between playlist and grouping of media on Android as well.

carolanitz:
> If the user imported a folder with a single file and the said file is deleted, the folder stays.

We need a follow up ticket to handle this case. Deletion with the files app is not an option since you force the user to leave our app and it's a "complicated workflow" 

> Please understand that there is a difference between playlist and grouping of media.

The whole comment was about users linking these things ;)  I'm very well aware of the distinction, and referred to the very first comment of the issue this PR is closing. (https://code.videolan.org/videolan/vlc-ios/issues/381)  ```the same folder structure should appear in VLC after uploading folder via WiFi sharing.```

I'm agreeing that it shouldn't be a playlist, as mentioned "pandoras box", but you need to bring this up and make a conscious decision and lay down reasons why this is not done or open tickets if necessary for follow up.

The reasoning with the albums and folder in folder is a comment that should be provided in 381 when closing this. folow up ticket for handling empty folders and potentially for media grouping and rest LGTM and can be merged 🎉

Mikanbu:
Rebased to master and addressed reviews.

PR #304, opened 2019-06-04
key command: add playback speed key commands

AndrewSB:
`]` to increase playback rate by 50%
`[` to decrease playback rate by 25%
`=` to reset the playback rate, only shown when the playback rate isn't 1


AndrewSB:
ping

would love to have this merged soon!

carolanitz:
@Mikanbu that can be changed in a followup as well I believe

carolanitz:
has been merged in with 484625492ec6a000da3fa509230288f46460583c 🎉

AndrewSB:
@Mikanbu I tried to keep it consistent with the macOS shortcuts :) 

If we should change the amount, update here and I'm happy to open a new PR

PR #307, opened 2019-06-06
WiFi Sharing: Thumbnail Fix

Mikanbu:
### Checklist

### Description

Closes .

### DIFF
--- a/Sources/VLCHTTPConnection.m
+++ b/Sources/VLCHTTPConnection.m
@@ -191,90 +191,68 @@
     return [allMedia copy];
 }
 
+- (NSString *)createHTMLMediaObjectFromMedia:(VLCMLMedia *)media
+{
+    return [NSString stringWithFormat:
+            @"<div style=\"background-image:url('Thumbnail/%@')\"> \
+            <a href=\"download/%@\" class=\"inner\"> \
+            <div class=\"down icon\"></div> \
+            <div class=\"infos\"> \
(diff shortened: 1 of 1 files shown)


Sources/VLCHTTPConnection.m
```diff
@@ -191,90 +191,68 @@ - (NSArray *)allMedia
...
+            </a> \
+            </div>",
+            media.thumbnail.absoluteString,
+            [[media mainFile].mrl.path
+             stringByAddingPercentEncodingWithAllowedCharacters:NSCharacterSet.URLFragmentAllowedCharacterSet],
+            media.title,
+            [media mediaDuration], [media formatSize]];
+}
+
+- (NSString *)createHTMLFolderObjectWithImagePath:(NSString *)imagePath
```
carolanitz: you could just use the VLCMLPlaylist object instead of three parameters but that's just a suggestion :)

Mikanbu: It was indeed my first implementation locally but since it was used by other types such as VLCMLAlbum, I decided to create it with specific parameters.

So since we are in Objective-C, we could either:
 - Use one global helper method
 - Use a "polymorphic" method that would be easier to use since we could just pass the classes directly

I didn't find the second solution "worth it" in this case since it was a minor addition. I believe it would be if it was used in more places or if it was, for example, a library API such as the medialibrary.

PR #308, opened 2019-06-06
Replace Sort & Edit by icons

Mikanbu:
### Checklist

### Description

Closes [#423](https://code.videolan.org/videolan/vlc-ios/issues/423).


Sources/MediaViewControllers/MediaViewController.swift
```diff
@@ -23,6 +23,7 @@ class VLCMediaViewController: VLCPagingViewController<VLCLabelCell>, MediaCatego
         self.services = services
         rendererButton = services.rendererDiscovererManager.setupRendererButton()
         super.init(nibName: nil, bundle: nil)
+        editButtonItem.customView = setupEditButton()
```
carolanitz: Didn't you want to just add an icon, is that not possible with UIImageview? otherwise I would go with an entire new button I just looked at the reference : https://developer.apple.com/documentation/uikit/uinavigationcontroller/customizing_your_app_s_navigation_bar

I think I would go with an entire new button to make clear that it is not the system editbuttonItem with it's system behavior.

Mikanbu:
Rebased and addressed review.

PR #321, opened 2019-06-24
MediaNavigationBar: Implementation of the new NavigationBar for MovieViewController on Figma

robwayne:
### Checklist


carolanitz:
linter violations ;)

robwayne:
> linter violations ;)

How do I test specifically for these? The fastlane tests are succeeding on my machine and building fine

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,135 @@
...
+    func didMinimizePlayback(_ navigationBar :MediaNavigationBar)
+    func didToggleChromeCast(_ navigationBar :MediaNavigationBar)
+    func didToggleAirPlay(_ navigationBar :MediaNavigationBar)
+}
+
+@objc (VLCMediaNavigationBar)
+@objcMembers class MediaNavigationBar :UIView {
+    
+    // MARK: Instance Variables
+    weak var delegate :VLCMediaNavigationBarDelegate?
```
carolanitz: in general we want ```var delegate: VLCMediaNavigationBarDelegate?``` [colon][space][class]

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,135 @@
...
+        return minButton
+    }()
+    
+    private var mediaTitleTextLabel :UILabel = {
+        var label = UILabel()
+        label.numberOfLines = 1
+        label.textColor = .white
+        label.font = UIFont(name: "SFProDisplay-Medium", size: 17)
+        label.text = NSLocalizedString("TITLE", comment: "Video Title")
+        label.isUserInteractionEnabled = false
```
carolanitz: why do you need to disable the isuserinteractionenabled?

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,135 @@
...
+        airPlayButton.setImage(UIImage(named: "iconAirplay"), for: .normal)
+        airPlayButton.tintColor = .white
+        airPlayButton.translatesAutoresizingMaskIntoConstraints = false
+        return airPlayButton
+    }()
+    
+    private func setupConstraints() {
+        NSLayoutConstraint.activate([
+            minimizePlaybackButton.leadingAnchor.constraint(equalTo: self.leadingAnchor),
+            minimizePlaybackButton.topAnchor.constraint(equalTo: self.topAnchor),
```
carolanitz: no self

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,135 @@
...
+    
+    private func setupViews() {
+        addSubview(minimizePlaybackButton)
+        addSubview(mediaTitleTextLabel)
+        addSubview(chromeCastButton)
+        addSubview(airPlayButton)
+        
+        if UIApplication.shared.statusBarOrientation.isLandscape {
+            mediaTitleTextLabel.isHidden = true
+        }
```
carolanitz: do we want that? if yes that needs to be handled on rotation as well

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,135 @@
...
+    }
+    
+    // MARK: Button Actions
+    func minimizePlayback() {
+        guard let delegate = delegate else {
+            print("Delegate not set for MediaNavigationBar")
+            return
+        }
+    
+        delegate.didMinimizePlayback(self)
```
carolanitz: it's simpler to write delegate?.didMinimizePlayback(self)

Sources/MediaNavigationBar.swift
robwayne: this was more of a preference to use self for inherited properties but I'll remove it

Sources/MediaNavigationBar.swift
robwayne: This was supposed to be Portrait and I'm not sure, the designs in figma are a bit contradictory. In a few the title is hidden on portrait mode and on others they are there. I just chose one

Sources/MediaNavigationBar.swift
robwayne: I wanted to at least make debugging down the line a lot easier, so that it would at least print the message to notify that the delegate was not set.

Sources/MediaNavigationBar.swift
carolanitz: okay in general you want to avoid it because then you know in swift code, everywhere where you have self it's potential for a retain cycle because only there you get forced to use self

Sources/MediaNavigationBar.swift
carolanitz: if it's just for debugging then we could also just add an assert(delegate != nil) :)
less code = more maintainable. Just as a suggestion but I can also live with the code above

Sources/MediaNavigationBar.swift
carolanitz: I'll take a look when I test it :)

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+ * Copyright © 2019 VLC authors and VideoLAN
+ *
+ * Authors: Robert Gordon <robwaynegordon # gmail.com>
+ *
+ *
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+@objc (VLCMediaNavigationBarDelegate)
+protocol VLCMediaNavigationBarDelegate {
```
Mikanbu: I believe you meant MediaNavigationBarDelegate for the protocol.

Sources/MediaNavigationBar.swift
robwayne: The navigation bar hasnt been added to the MovieViewController as yet, that would be in the next PR

robwayne:
Re-open for review @carolanitz

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+        airPlayButton.tintColor = .white
+        airPlayButton.translatesAutoresizingMaskIntoConstraints = false
+        return airPlayButton
+    }()
+    
+    private func setupConstraints() {
+        NSLayoutConstraint.activate([
+            minimizePlaybackButton.leadingAnchor.constraint(equalTo: leadingAnchor),
+            minimizePlaybackButton.topAnchor.constraint(equalTo: topAnchor),
+            minimizePlaybackButton.widthAnchor.constraint(equalToConstant: minimizePlaybackButton.frame.width),
```
carolanitz: this shouldn't be needed. It should resize itself based on the image

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+            minimizePlaybackButton.bottomAnchor.constraint(equalTo: bottomAnchor),
+            mediaTitleTextLabel.leadingAnchor.constraint(equalTo: minimizePlaybackButton.trailingAnchor, constant: 16),
+            mediaTitleTextLabel.topAnchor.constraint(equalTo: topAnchor),
+            mediaTitleTextLabel.trailingAnchor.constraint(equalTo: chromeCastButton.leadingAnchor, constant: 50),
```
carolanitz: I wouldn't define hard the chromecast button in relation to the mediatitle because the spacing between the two can be variable big. I'd just define lessthan constraint

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+            chromeCastButton.trailingAnchor.constraint(equalTo: airPlayButton.leadingAnchor, constant: 20),
```
carolanitz: according to figma it's 33 or 35 depending on how you measure. but 20 is fine and I would also use the same distance between chevron and title

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+            chromeCastButton.topAnchor.constraint(equalTo: topAnchor),
+            chromeCastButton.widthAnchor.constraint(equalToConstant: 24),
```
carolanitz: width not necessary because it should resize itself by the image

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+            airPlayButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: 20),
```
carolanitz: if the trailing padding at the end is 20 it should also be 20 at the beginning and since we now have 20 everywhere I would make it a let margin = 20 and replace it everywhere

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+        ])
+    }
+    
+    private func setupViews() {
+        addSubview(minimizePlaybackButton)
+        addSubview(mediaTitleTextLabel)
+        addSubview(chromeCastButton)
+        addSubview(airPlayButton)
+        
+        mediaTitleTextLabel.isHidden = UIApplication.shared.statusBarOrientation.isPortrait
```
carolanitz: this is actually an inconsistency in the design. the title should always be there and indeed you can see that if you look at all designs that it's sometimes there and sometimes not

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+        return chromeButton
+    }()
+    
+    lazy var airPlayButton: UIButton = {
+        var airPlayButton = UIButton(type: .system)
+        airPlayButton.addTarget(self, action: #selector(toggleAirPlay), for: .touchUpInside)
+        airPlayButton.setImage(UIImage(named: "iconAirplay"), for: .normal)
+        airPlayButton.tintColor = .white
+        airPlayButton.translatesAutoresizingMaskIntoConstraints = false
+        return airPlayButton
```
carolanitz: I believe there is a system component called AVRoutepickerview that needs to be used here instead https://developer.apple.com/documentation/avkit/avroutepickerview

but I'm not sure if that will work with our openGL player

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+@objcMembers class MediaNavigationBar: UIView {
+    
+    // MARK: Instance Variables
+    weak var delegate: MediaNavigationBarDelegate?
+    
+    var hideTitleLabel: Bool {
+        didSet {
+            mediaTitleTextLabel.isHidden = hideTitleLabel
+        }
+    }
```
carolanitz: shouldn't be hidden and always visible

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,128 @@
...
+        minButton.tintColor = .white
+        minButton.translatesAutoresizingMaskIntoConstraints = false
+        return minButton
+    }()
+    
+    private var mediaTitleTextLabel: UILabel = {
+        var label = UILabel()
+        label.numberOfLines = 1
+        label.textColor = .white
+        label.font = UIFont(name: "SFProDisplay-Medium", size: 17)
```
carolanitz: systemfont. the systemfont is sfprodisplay by default but by using that one we will have no trouble should apple change it in the future to something else

Sources/MediaNavigationBar.swift
robwayne: This is only available in xcode 11, should I just implement it without fallback?

carolanitz: ugh fuuuu okay yeah I would implement it with fallback and just if #available or something like that. That way we can just drop the legacy fun when we drop ios 9 and 10 one day

carolanitz:
I believe you want to merge two of the commits, because they have the same title and message?

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,134 @@
...
+ * Refer to the COPYING file of the official project for license.
+ *****************************************************************************/
+
+import AVKit
+
+@objc (VLCMediaNavigationBarDelegate)
+protocol MediaNavigationBarDelegate {
+    func didMinimizePlayback(_ navigationBar: MediaNavigationBar)
+    func didToggleChromeCast(_ navigationBar: MediaNavigationBar)
+    func didToggleAirPlay(_ navigationBar: MediaNavigationBar)
```
carolanitz: naming convention is rather medianavigationbarDidToggleChromecast(_ medianavigationbar: MediaNavigationBar) i believe see https://developer.apple.com/documentation/uikit/uiscrollviewdelegate

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,134 @@
...
+}
+
+@objc (VLCMediaNavigationBar)
+@objcMembers class MediaNavigationBar: UIView {
+    
+    // MARK: Instance Variables
+    weak var delegate: MediaNavigationBarDelegate?
```
carolanitz: 👍for the weak

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,134 @@
...
+    }()
+    
+    @available(iOS 11.0, *)
+    lazy var airPlayRoutePickerViewButton: AVRoutePickerView = {
+        var airPlayRoutePicker = AVRoutePickerView(frame: .zero)
+        airPlayRoutePicker.activeTintColor = .orange
+        airPlayRoutePicker.tintColor = .white
+        airPlayRoutePicker.translatesAutoresizingMaskIntoConstraints = false
+        return airPlayRoutePicker
+    }()
```
carolanitz: yeah just return different buttons here depending on the os version

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,134 @@
...
+            airplayConstraints = [
+                chromeCastButton.trailingAnchor.constraint(equalTo: trailingAnchor, constant: margin)
+            ]
+        }
+        
+        let constraints: [NSLayoutConstraint] = [
+            minimizePlaybackButton.leadingAnchor.constraint(equalTo: leadingAnchor),
+            minimizePlaybackButton.topAnchor.constraint(equalTo: topAnchor),
+            minimizePlaybackButton.bottomAnchor.constraint(equalTo: bottomAnchor),
+            mediaTitleTextLabel.leadingAnchor.constraint(equalTo: minimizePlaybackButton.trailingAnchor, constant: margin),
```
carolanitz: +1 for the margin variable

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,134 @@
...
+            mediaTitleTextLabel.topAnchor.constraint(equalTo: topAnchor),
+            mediaTitleTextLabel.trailingAnchor.constraint(lessThanOrEqualTo: chromeCastButton.leadingAnchor),
```
carolanitz: you probably want to add the margin here as well

Sources/MediaNavigationBar.swift
```diff
@@ -0,0 +1,134 @@
...
+             addSubview(airPlayRoutePickerViewButton)
+        }
+    }
+    
+    // MARK: Initializers
+    required init(coder: NSCoder) {
+        fatalError("init(coder: NSCoder) not implemented")
+    }
+    
+    init() {
```
carolanitz: overwrite the designated initializer here instead and call it with  .zero from movieviewcontroller later

Sources/MediaNavigationBar.swift
robwayne: "ugh fuuuu" hahahah 😂okay done

Sources/MediaNavigationBar.swift
robwayne: thanks for the extra credit haha

Sources/MediaNavigationBar.swift
robwayne: Should I return a regular UIButton?

robwayne:
will this now be merged?

PR #322, opened 2019-06-24
VLCPlaybackController: only restore playbackposition when less than 9…

carolanitz:
…5% were viewed

Otherwise a user might be unable to rewatch content because it would immediatly close the controller again

### Checklist

### Description

closes #522


carolanitz:
was merged with c3f4d8e40055811fe5ede40e2f3957f2b573736a

PR #325, opened 2019-06-25
EditController: Playlist: Handle collection addition

Mikanbu:
### Checklist

### Description

Quick and straight forward patch that adds the possibility to add any collection(artist, album, genre, ...) to a playlist.

### DIFF
--- a/Sources/EditController.swift
+++ b/Sources/EditController.swift
@@ -332,8 +332,18 @@ extension EditController: AddToPlaylistViewControllerDelegate {
     func addToPlaylistViewController(_ addToPlaylistViewController: AddToPlaylistViewController,
                                      didSelectPlaylist playlist: VLCMLPlaylist) {
         let files = model.anyfiles
+        var mediaObjects = [VLCMLObject]()
+
         for index in selectedCellIndexPaths where index.row < files.count {
-            if !playlist.appendMedia(withIdentifier: files[index.row].identifier()) {
+            if let mediaCollection = files[index.row] as? MediaCollectionModel {
+                mediaObjects += mediaCollection.files() ?? []
+            } else {
+                mediaObjects.append(files[index.row])
+            }
+        }
+
+        for media in mediaObjects {
+            if !playlist.appendMedia(withIdentifier: media.identifier()) {
                 assertionFailure("EditController: AddToPlaylistViewControllerDelegate: Failed to add item.")
             }
         }


Sources/EditController.swift
```diff
@@ -319,9 +319,23 @@ extension EditController: AddToPlaylistViewControllerDelegate {
...
+
+                for file in collectionFiles {
+                    if !playlist.appendMedia(withIdentifier: file.identifier()) {
+                        assertionFailure("EditController: AddToPlaylistViewControllerDelegate: Failed to add item.")
+                    }
+                }
+            } else {
+                if !playlist.appendMedia(withIdentifier: files[index.row].identifier()) {
+                    assertionFailure("EditController: AddToPlaylistViewControllerDelegate: Failed to add item.")
+                }
```
Mikanbu: Tiny duplication, if it is a deal breaker, I can look into it later on if needed.

_I need to go back to medialibrary cell states coding._ 🤐

Sources/EditController.swift
```diff
@@ -319,9 +319,23 @@ extension EditController: AddToPlaylistViewControllerDelegate {
...
         let files = model.anyfiles
+
         for index in selectedCellIndexPaths where index.row < files.count {
-            if !playlist.appendMedia(withIdentifier: files[index.row].identifier()) {
-                assertionFailure("EditController: AddToPlaylistViewControllerDelegate: Failed to add item.")
+            if let mediaCollection = files[index.row] as? MediaCollectionModel {
+                guard let collectionFiles = mediaCollection.files() else {
+                    // unable to retreive files from a media collection
+                    continue
+                }
```
carolanitz: I don't think you need the guard here, you can just say ```for file in mediaCollection.files()``` if it's nil it will just not jump into the for loop, which has the same effect as the continue but I guess the shortest is something like this
```
let mediaCollection = files[index.row] as? MediaCollectionModel
let filesToAppend = mediaCollection != nil ? mediaCollection.files() : [files[index.row]]
for file in filesToAppend {
    if !playlist.appendMedia(withIdentifier: file.identifier()) {
        assertionFailure("EditController: AddToPlaylistViewControllerDelegate: Failed to add item.")
    }
}

Mikanbu:
Updated to remove duplication.

Mikanbu:
Rebased to master.

PR #330, opened 2019-06-26
AppCoordinator: Fix invisible 6th tab in TabBar

carolanitz:
initializing the tabbarcoordinator before we setup the childViewControllers leads to a 6th tab
Likely caused by tabBarController.addChild(playerController) which contains our miniplayer

instantiating the tabbarcoordinator lazily fixes it
(closes #539)

### Checklist


carolanitz:
I'll write up a ticket that we should fix the view hierarchy and not make the playerController a child of the tabbar but right now I wouldn't want to have changes so high up in our hierarchy

PR #335, opened 2019-06-28
Fastlane app content

carolanitz:
### Checklist

### Description

The general dummy data and the appdata to test this PR with is herehttps://drive.google.com/open?id=1TJhourtjOx8kztMl3PR8o4hjl_zNJqpE

For testing I adjust the snapfile to only contain one device and one language and I disable the menu tests.
We probably need even more data but this but it's a start, and as a follow up we should fix which testssuites should run for snapshotting and which for actual testing on CI and not run both in both cases :D


carolanitz:
guess I need to fix the testsuite rather now as part of this PR

carolanitz:
merged with c1a7949be2d0fcb89a2a1e44227597c5a12252fd & previous

PR #339, opened 2019-07-08
VLCMetadata: Show data in Miniplayer even when passcode is enabled

carolanitz:
Before we would immediatly return, now we return before populating the
infocenter

closes #511

### Checklist

### DIFF
--- a/SharedSources/VLCMetadata.m
+++ b/SharedSources/VLCMetadata.m
@@ -43,7 +43,6 @@
 #if TARGET_OS_IOS
 - (void)updateMetadataFromMedia:(VLCMLMedia *)media mediaPlayer:(VLCMediaPlayer*)mediaPlayer
 {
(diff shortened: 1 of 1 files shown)


SharedSources/VLCMetadata.m
```diff
@@ -61,6 +60,9 @@ - (void)updateMetadataFromMedia:(VLCMLMedia *)media mediaPlayer:(VLCMediaPlayer*
     }
     [self updatePlaybackRate:mediaPlayer];
```
Mikanbu: I'm not sure but do we need to `updatePlaybackRate` if the passcode lock is enabled? 🤔

carolanitz: tbh I don't think so but I would not move the check above that line because it's more readable / logical to just have it above populateInfoCenter.

carolanitz:
merged with 91184d4a261655a7c0c214b325ac4056d6c145a5

PR #347, opened 2019-07-14
Extend ActionSheetCell: Replace checkmark label with class property ActionSheetAccessoryType

robwayne:
Replaced the previous checkmark label and updated it to utilize a property similar to UICollectionViewCell's accessoryType.

### Checklist


Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -9,6 +9,58 @@
...
+        case .none:
+            switch rhs {
+            case .none:
+                return true
+            default:
+                return false
+            }
+        default:
+            assertionFailure("Unhandled ActionSheetAccessoryType")
+        }
```
carolanitz: I'm pretty sure there is a better way than this! Can you spend some more time researching? 
Additionally we like to avoid default to explicitly handle defined cases

Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -9,6 +9,58 @@
...
+
+extension UIImage {
+    class var checkmark: UIImage? {
+        return UIImage(named: "checkmark")?.withRenderingMode(.alwaysTemplate)
+    }
+    
+    class var disclosureChevron: UIImage? {
+        return UIImage(named: "disclosureChevron")?.withRenderingMode(.alwaysTemplate)
+    }
+}
```
carolanitz: this is a project wide extension of UIImage. I feel a little bit uncomfortable with this, can we somehow limit this to this class and the classes that interact with it?

Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -56,16 +107,35 @@ class ActionSheetCell: UICollectionViewCell {
...
+            case .disclosureChevron:
+                accessoryTypeImageView.image = .disclosureChevron
+            case .custom(let image):
+                accessoryTypeImageView.image = image.withRenderingMode(.alwaysTemplate)
+            case .none:
+                accessoryTypeImageView.image = nil
+                accessoryTypeImageView.isHidden = true
+            default:
+                assertionFailure("Unhandled ActionSheetCellAccessoryType")
+            }
```
carolanitz: Are these cells being reused? if so you could end up with a hidden imageview if you don't set that correctly in the other cases

Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -9,6 +9,58 @@
  * Refer to the COPYING file of the official project for license.
  *****************************************************************************/
 
+enum ActionSheetCellAccessoryType: Equatable {
+    case none
+    case checkmark
+    case disclosureChevron
+    case custom(image: UIImage)
```
carolanitz: can we remove the custom? General rule of thumb: Don't try to make something generic before it's needed since it makes code often more complicated

Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -56,16 +107,35 @@ class ActionSheetCell: UICollectionViewCell {
...
+        let imageView = UIImageView()
+        imageView.contentMode = .scaleAspectFit
+        imageView.backgroundColor = .none
+        imageView.tintColor = PresentationTheme.current.colors.cellDetailTextColor
+        imageView.image = .checkmark
+        imageView.translatesAutoresizingMaskIntoConstraints = false
+        return imageView
     }()
+    
+    var accessoryType: ActionSheetCellAccessoryType {
```
carolanitz: you can also assign a first value var accessoryType = .checkmark { ...

Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -77,11 +147,13 @@ class ActionSheetCell: UICollectionViewCell {
     }()
 
     override init(frame: CGRect) {
+        accessoryType = .checkmark
         super.init(frame: frame)
         setupViews()
     }
 
     required init?(coder aDecoder: NSCoder) {
+        accessoryType = .checkmark
```
carolanitz: if you assign the .checkmark as default you don't need those

Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -90,6 +162,11 @@ class ActionSheetCell: UICollectionViewCell {
         let colors = PresentationTheme.current.colors
         name.textColor = isSelected ? colors.orangeUI : colors.cellTextColor
         tintColor = isSelected ? colors.orangeUI : colors.cellDetailTextColor
+        if accessoryType == .checkmark {
+            accessoryTypeImageView.isHidden = !isSelected
```
carolanitz: isHidden has nothing to do with updateColors ^ that can be removed I believe.
We need to double-check here with Louis if the action sheet should always be darktheme styled in the movieviewcontroller, if so we need to make sure that this gets correctly called in the chevron case

Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -98,11 +175,13 @@ class ActionSheetCell: UICollectionViewCell {
     }
 
     private func setupViews() {
+        accessoryType = .checkmark // set the accessoryType initial value again so that the property observer didSet is called
+
```
carolanitz: I wonder if this is still needed if you set a default

Sources/ActionSheet/ActionSheetCell.swift
Mikanbu: Can't this be removed? I'm not sure this overload is needed in this case.

Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -56,16 +107,35 @@ class ActionSheetCell: UICollectionViewCell {
...
-
-    let checkmark: UILabel = {
-        let checkmark = UILabel()
-        checkmark.text = "✓"
-        checkmark.font = UIFont.systemFont(ofSize: 18)
-        checkmark.textColor = PresentationTheme.current.colors.orangeUI
-        checkmark.translatesAutoresizingMaskIntoConstraints = false
-        checkmark.isHidden = true
-        return checkmark
+
```
Mikanbu: Nit: Extra additions.

Sources/ActionSheet/ActionSheetCell.swift
robwayne: I removed it in the end and just used regular UIImage(named:)

Sources/ActionSheet/ActionSheetCell.swift
robwayne: this will only trigger for the checkmark options so the colors of the chevron will always be the same tint of grey, in dark and light themes

Sources/ActionSheet/ActionSheetCell.swift
```diff
@@ -56,16 +63,33 @@ class ActionSheetCell: UICollectionViewCell {
...
+                accessoryTypeImageView.image = UIImage(named: "checkmark")?.withRenderingMode(.alwaysTemplate)
+                accessoryTypeImageView.isHidden = !isSelected
+            case .disclosureChevron:
+                accessoryTypeImageView.image = UIImage(named: "disclosureChevron")?.withRenderingMode(.alwaysTemplate)
+                accessoryTypeImageView.isHidden = false
+            case .none:
+                accessoryTypeImageView.image = nil
+                accessoryTypeImageView.isHidden = true
+            default:
+                assertionFailure("Unhandled ActionSheetCellAccessoryType")
```
Mikanbu: Swift compiler complaining here.

robwayne:
> I mostly agree with this change but I'm wondering on the usage part.
> 
> Should we consider a way of customizing the accessory within an initializer?

Yeah I also was considering that as well, but since most are going to be dequeued and reused anyways will we be initializing many of them in anycase?

Sources/ActionSheet/ActionSheetCell.swift
robwayne: @carolanitz says its best to keep the default as an assertionFailure so that if we add more cases that we dont catch, it will assert fail

Mikanbu: In the case where we remove the default and add more cases to the enum without adding it to in the switch, Swift will most likely complain saying something like: `switch must be exhaustive...`

Mikanbu:
Valid point, should we consider it to be an init to ActionSheet then? 🤔

Sources/ActionSheet/ActionSheetCell.swift
robwayne: Yeah that is true but also I also agree with Caro because although not entirely likely, what happens if the compiler is changed and that is removed. I don't have strong feelings either way, I as well dont like seeing the extra warnings so I can do either, if you have stronger feelings I dont mind changing it

robwayne:
> Valid point, should we consider it to be an init to ActionSheet then? 🤔

Yeah I like this, that's something that we could do. Something like a `ActionSheet(withAccessoryType:)` correct? If so should I add another convenient initializer or add it to the other current inits: `init(header:)`/ `init()`

Mikanbu:
I would say another convenience init.

Anyway, for me before we merge this there are two small things that needs to be corrected/decided on:

- Switch default case
- Nit-pick: extra spaces l102

robwayne:
> I would say another convenience init.
> 
> Anyway, for me before we merge this there are two small things that needs to be corrected/decided on:
> 
> * Switch default case
> * Nit-pick: extra spaces l102

Okay. Done, also after checking it out more I realized the initializer for the ActionSheet would also be a bit impractical because since the cells are determined by the actionsheet's delegate and the cells are reused as well they would have to be set in the cellforindexpath

Mikanbu:
Merged with 02b7d6b609a34994c709debd6785f2b050ee6c2f and 5da28d4c59ba42060e36cd126c267c854e74b1a8

PR #348, opened 2019-07-14
Tx translations from 15.7.

mtrz:
Only 100% complete translations of the last strings


carolanitz:
merged with ae2e828cd24aadb7d2934e1ac23de5abd71d3cbb and following

PR #350, opened 2019-07-18
VideoOptionsControlBar: Update toggle fullscreen button image

robwayne:
### Checklist

### DIFF
--- a/Sources/VLCMovieViewController.m
+++ b/Sources/VLCMovieViewController.m
@@ -1159,6 +1159,8 @@ typedef NS_ENUM(NSInteger, VLCPanType) {
 - (void)playbackDidStop:(NSNotification *)notification
 {
     [self minimizePlayback:nil];
+    // reset toggleButton to default icon when movieviewcontroller is dismissed
+    _videoOptionsControlBar.isInFullScreen = NO;
 }
 
 - (void)mediaPlayerStateChanged:(VLCMediaPlayerState)currentState
@@ -1206,6 +1208,11 @@ currentMediaHasTrackToChooseFrom:(BOOL)currentMediaHasTrackToChooseFrom
                      }];
 }
 
+- (void) playbackControllerDidSwitchAspectRatio:(NSUInteger)aspectRatio
+{
+    _videoOptionsControlBar.isInFullScreen = aspectRatio == VLCAspectRatioFillToScreen;
+}
+
 - (void)displayMetadataForPlaybackController:(VLCPlaybackController *)controller metadata:(VLCMetaData *)metadata
 {
     if (!_viewAppeared)
@@ -1222,8 +1229,12 @@ currentMediaHasTrackToChooseFrom:(BOOL)currentMediaHasTrackToChooseFrom
     [self hideShowAspectratioButton:metadata.isAudioOnly];
     [_controllerPanel updateButtons];
     [_playingExternalView updateUIWithRendererItem:_vpc.renderer title:metadata.title];
-    
+
     _audioOnly = metadata.isAudioOnly;
+
+    #if NEW_UI
+        _videoOptionsControlBar.toggleFullScreenButton.hidden = _audioOnly;
+    #endif
 }
 
 - (IBAction)playPause
@@ -1669,7 +1680,7 @@ currentMediaHasTrackToChooseFrom:(BOOL)currentMediaHasTrackToChooseFrom
     } else if (tapPosition.x > forwardBoundary){
         _numberOfTapSeek = _previousJumpState == VLCMovieJumpStateBackward ? 1 : _numberOfTapSeek + 1;
     } else {
-        [_vpc toggleFullScreen];
+        [_vpc switchAspectRatio:YES];
         return;
     }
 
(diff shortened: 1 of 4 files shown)


robwayne:
@carolanitz @Mikanbu ping!

Sources/VLCPlaybackController.m
```diff
@@ -57,6 +57,9 @@ @interface VLCPlaybackController () <VLCMediaPlayerDelegate, VLCMediaDelegate, V
 
     NSUInteger _currentAspectRatio;
     BOOL _isInFillToScreen;
+    BOOL _toggledFullScreen;
+    NSUInteger _previousAspectRatio;
```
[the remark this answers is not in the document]
robwayne: Currently, it's necessary to accommodate the current (old) ui. The changeAspectRatio button toggles between different video ratios, but additionally you can enter directly into fullScreen by double tapping, therefore knowing when a request for fullScreen mode is requested is necessary

Sources/VLCMovieViewController.m
```diff
@@ -1205,6 +1207,12 @@ - (void)hideShowAspectratioButton:(BOOL)hide
                      }];
 }
 
+- (void) playbackControllerDidSwitchAspectRatio:(NSUInteger)aspectRatio
+{
+    // if aspectRatio is VLCAspectRatioFillToScreen
+    _videoOptionsControlBar.isInFullScreen = aspectRatio == VLCAspectRatioFillToScreen;
```
carolanitz: you don't need that comment, that is already in the code

Sources/VideoOptionsControlBar.swift
```diff
@@ -121,8 +128,6 @@ protocol VideoOptionsControlBarDelegate: class {
     
     override init(frame: CGRect) {
         repeatMode = .doNotRepeat
-        subtitleToggled = false
-        interfaceDisabled = false
```
carolanitz: why was that removed as part of toggle fullscreen button image?

Sources/VideoOptionsControlBar.swift
```diff
@@ -24,13 +24,20 @@ protocol VideoOptionsControlBarDelegate: class {
...
+    }
+    
+    var subtitleToggled: Bool = false {
         didSet {
             selectSubtitleButton.tintColor = subtitleToggled ? .orange : .white
         }
     }
     
-    var interfaceDisabled: Bool {
+    var interfaceDisabled: Bool = false {
```
carolanitz: a bool is always already false when it's initialized

Sources/VLCPlaybackController.h
```diff
@@ -42,6 +50,7 @@ currentMediaHasTrackToChooseFrom:(BOOL)currentMediaHasTrackToChooseFrom
 - (void)prepareForMediaPlayback:(VLCPlaybackController *)controller;
 - (void)showStatusMessage:(NSString *)statusMessage;
 - (void)displayMetadataForPlaybackController:(VLCPlaybackController *)controller metadata:(VLCMetaData *)metadata;
+- (void)playbackControllerDidSwitchAspectRatio:(NSUInteger)aspectRatio;
```
carolanitz: VLCAspectRatio not NSUInteger

Sources/VideoOptionsControlBar.swift
robwayne: Swift complains about initialization since its a property observer

Sources/VideoOptionsControlBar.swift
robwayne: Set their default values as false

Sources/VLCMovieViewController.m
```diff
@@ -564,6 +564,8 @@ - (void)viewWillDisappear:(BOOL)animated
 
 - (void)viewDidDisappear:(BOOL)animated
 {
+    // reset toggleButton to default icon when movieviewcontroller is dismissed
+    _videoOptionsControlBar.isInFullScreen = NO;
```
carolanitz: that leads to a wrong state, if you go into fullscreen and switch to miniplayer and back to maxiplayer the button has the wrong state because of this

Sources/VLCPlaybackController.m
```diff
@@ -834,6 +833,8 @@ - (void)switchAspectRatio
     if ([self.delegate respondsToSelector:@selector(showStatusMessage:)]) {
         [self.delegate showStatusMessage:[NSString stringWithFormat:NSLocalizedString(@"AR_CHANGED", nil), [self stringForAspectRatio:_currentAspectRatio]]];
     }
+
+    [_delegate playbackControllerDidSwitchAspectRatio:_currentAspectRatio];
```
carolanitz: you need to make a check here if the delegate responds to the selector, that call is under an @optional which means that a delegate doesn't need to implement it. Even though you might have a delegate this could crash when the function is not implemented

Sources/VLCMovieViewController.m
carolanitz: I would set that in prepareForMediaPlayback which is only called when a new media is started

Sources/VLCPlaybackController.m
robwayne: Okay, ahh thats what that does

Sources/VLCMovieViewController.m
robwayne: Actually, prepareForMediaPlayback is called everytime the view comes back into view. But I didnt know that there was a playbackDidStop notification, which is better for resetting I believe.

PR #352, opened 2019-07-19
VLCMovieViewController: Integrate MediaNavigationBar

robwayne:
### Checklist

### DIFF
--- a/Resources/en.lproj/Localizable.strings
+++ b/Resources/en.lproj/Localizable.strings
@@ -371,7 +371,7 @@
 "VARIOUS_ARTIST" = "Various Artists";
 "UNKNOWN_ALBUM" = "Unknown Album";
 
-/* New strings */
+/* VideoOptionsControlBar */
 "INTERFACE_LOCK_BUTTON" = "Lock interface";
 "INTERFACE_LOCK_HINT" = "Disable interface controls";
 "MORE_OPTIONS_BUTTON" = "More";
--- a/Sources/MediaNavigationBar.swift
+++ b/Sources/MediaNavigationBar.swift
@@ -13,8 +13,9 @@ import AVKit
 
 @objc (VLCMediaNavigationBarDelegate)
 protocol MediaNavigationBarDelegate {
-    func mediaNavigationBarDidMinimizePlayback(_ navigationBar: MediaNavigationBar)
-    func mediaNavigationBarDidToggleChromeCast(_ navigationBar: MediaNavigationBar)
+    func mediaNavigationBarDidTapMinimize(_ mediaNavigationBar: MediaNavigationBar)
+    func mediaNavigationBarDidLongPressMinimize(_ mediaNavigationBar: MediaNavigationBar)
+    func mediaNavigationBarDidToggleChromeCast(_ mediaNavigationBar: MediaNavigationBar)
 }
 
 @objc (VLCMediaNavigationBar)
@@ -22,24 +23,27 @@ protocol MediaNavigationBarDelegate {
     
     // MARK: Instance Variables
     weak var delegate: MediaNavigationBarDelegate?
+    private var buttonSize: CGFloat = 24
     
     lazy var minimizePlaybackButton: UIButton = {
         var minButton = UIButton(type: .system)
-        minButton.addTarget(self, action: #selector(minimizePlayback), for: .touchUpInside)
+        let longPressGesture = UILongPressGestureRecognizer(target: self,
+                                action: #selector(handleMinimizeLongPress))
+        minButton.addGestureRecognizer(longPressGesture)
+        minButton.addTarget(self, action: #selector(handleMinimizeTap), for: .touchUpInside)
         minButton.setImage(UIImage(named: "iconChevron"), for: .normal)
         minButton.tintColor = .white
-        minButton.translatesAutoresizingMaskIntoConstraints = false
+        minButton.setContentHuggingPriority(.defaultHigh, for: .horizontal)
         return minButton
     }()
     
-    private var mediaTitleTextLabel: UILabel = {
+    lazy var mediaTitleTextLabel: UILabel = {
         var label = UILabel()
         label.numberOfLines = 1
-        label.lineBreakMode = .byTruncatingTail
+        label.lineBreakMode = .byTruncatingMiddle
         label.textColor = .white
         label.font = UIFont(name: "SFProDisplay-Medium", size: 17)
-        label.text = NSLocalizedString("TITLE", comment: "")
-        label.translatesAutoresizingMaskIntoConstraints = false
+        label.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
         return label
     }()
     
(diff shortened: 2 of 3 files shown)


Sources/MediaNavigationBar.swift
```diff
@@ -77,59 +88,41 @@ protocol MediaNavigationBarDelegate {
...
-        ]
-        
-        NSLayoutConstraint.activate(constraints)
-    }
-    
     private func setupViews() {
-        addSubview(minimizePlaybackButton)
-        addSubview(mediaTitleTextLabel)
-        addSubview(chromeCastButton)
+        axis = .horizontal
```
Mikanbu: `UILayoutConstraintAxisHorizontal` should already be the default on UIStackViews.

Sources/MediaNavigationBar.swift
```diff
@@ -77,59 +88,41 @@ protocol MediaNavigationBarDelegate {
...
+        spacing = 20.0
+        alignment = .fill
```
Mikanbu: Same remark as the axis.

Sources/MediaNavigationBar.swift
```diff
@@ -77,59 +88,41 @@ protocol MediaNavigationBarDelegate {
...
+        assert(delegate != nil, "Delegate not set for MediaNavigationBar")
+        delegate?.mediaNavigationBarDidLongPressMinimize(self)
     }
     
     func toggleChromeCast() {
         assert(delegate != nil, "Delegate not set for MediaNavigationBar")
         delegate?.mediaNavigationBarDidToggleChromeCast(self)
-    }
-}
+    }}
```
Mikanbu: 🤔

Resources/en.lproj/Localizable.strings
```diff
@@ -378,3 +378,8 @@
 "MORE_OPTIONS_HINT" = "View more option controls";
 "REPEAT_MODE_HINT" = "Change repeat mode of current item";
 "VIDEO_ASPECT_RATIO_HINT" = "Change the aspect ratio of the current video";
+
+/* New strings */
+
+/* New strings */
+"MEDIA_NAV_TITLE" = "Media Title";
```
Mikanbu: I would remove `/* New strings */` or perhaps replace it by the used class/context?

Sources/VLCMovieViewController.m
```diff
@@ -44,7 +45,7 @@
 #define DEFAULT_FOV 80.f
 #define MAX_FOV 150.f
 #define MIN_FOV 20.f
-#define NEW_UI 0
+#define NEW_UI 1
```
Mikanbu: I don't think this change should be committed.
Additionally, it seems that it doesn't compile with the "old" UI.

Resources/en.lproj/Localizable.strings
robwayne: used class context?

Mikanbu: In this case it would be `MediaNavigationBar`

robwayne: oh haha u mean the comment not the localized string itself. I need to learn to read lol. Okay done!

robwayne:
ping!

Sources/MediaNavigationBar.swift
```diff
@@ -57,6 +66,7 @@ protocol MediaNavigationBarDelegate {
         var airPlayRoutePicker = AVRoutePickerView(frame: .zero)
         airPlayRoutePicker.activeTintColor = .orange
         airPlayRoutePicker.tintColor = .white
+        airPlayRoutePicker.widthAnchor.constraint(equalToConstant: buttonSize).isActive = true
```
carolanitz: is this really necessary? don't the buttons resize themself based on the image?

Sources/VLCMovieViewController.m
carolanitz: this should be 0 I agree

Sources/VLCMovieViewController.m
```diff
@@ -54,7 +55,7 @@ typedef NS_ENUM(NSInteger, VLCPanType) {
   VLCPanTypeProjection
 };
 
-@interface VLCMovieViewController () <UIGestureRecognizerDelegate, VLCMultiSelectionViewDelegate, VLCEqualizerViewUIDelegate, VLCPlaybackControllerDelegate, VLCDeviceMotionDelegate, VLCRendererDiscovererManagerDelegate, PlaybackSpeedViewDelegate, VLCVideoOptionsControlBarDelegate>
+@interface VLCMovieViewController () <UIGestureRecognizerDelegate, VLCMultiSelectionViewDelegate, VLCEqualizerViewUIDelegate, VLCPlaybackControllerDelegate, VLCDeviceMotionDelegate, VLCRendererDiscovererManagerDelegate, PlaybackSpeedViewDelegate, VLCVideoOptionsControlBarDelegate, VLCMediaNavigationBarDelegate>
```
carolanitz: you know what? Looking at all the code that gets added I would add a Swift extension for MovieViewController that includes all the delegate methods for the new UI and other new UI code to have clearer distinguishment between new and old

Sources/VLCMovieViewController.m
```diff
@@ -1383,7 +1435,11 @@ - (void)toggleUILock
     for (NSObject *item in items) {
         if ([item isKindOfClass:[UIControl class]]) {
             UIControl *control = (UIControl *)item;
-            control.enabled = !_interfaceIsLocked;
+            if (NEW_UI && item == _mediaNavigationBar.chromeCastButton) {
+                control.enabled = !_interfaceIsLocked && _vpc.renderer != nil;
```
carolanitz: that looks wrong. The UILock should be not effected by having a renderer

Sources/MediaNavigationBar.swift
```diff
@@ -48,7 +55,9 @@ protocol MediaNavigationBarDelegate {
         chromeButton.addTarget(self, action: #selector(toggleChromeCast), for: .touchUpInside)
         chromeButton.setImage(UIImage(named: "renderer"), for: .normal)
         chromeButton.tintColor = .white
+        chromeButton.widthAnchor.constraint(equalToConstant: buttonSize).isActive = true
         chromeButton.translatesAutoresizingMaskIntoConstraints = false
+        chromeButton.isEnabled = false // disable until a chromecast device is found
```
carolanitz: that button should not be visible I assume when there is no chromecast device

Sources/MediaNavigationBar.swift
```diff
@@ -13,32 +13,39 @@ import AVKit
...
     
-    private var mediaTitleTextLabel: UILabel = {
+    lazy var mediaTitleTextLabel: UILabel = {
         var label = UILabel()
         label.numberOfLines = 1
         label.lineBreakMode = .byTruncatingTail
         label.textColor = .white
         label.font = UIFont(name: "SFProDisplay-Medium", size: 17)
-        label.text = NSLocalizedString("TITLE", comment: "Video Title")
+        label.text = NSLocalizedString("MEDIA_NAV_TITLE", comment: "")
```
carolanitz: I would say this needs to be empty unless we have a media title?

Sources/MediaNavigationBar.swift
robwayne: hmm would it not be just as okay to have the button but not have it selectable if a renderer isnt present?

Sources/VLCMovieViewController.m
robwayne: This is actually setting the chromecast button to be enabled only if there is a renderer device and the interface lock is not locked. aka if the interface is unlocked but no renderer it remains disabled

Sources/MediaNavigationBar.swift
robwayne: Actually despite trying setContentHuggingPriority on the buttons, the minimize and chromecast buttons are unwilling to stay up without the widths set. "Width and horizontal position are ambiguous"

Sources/VLCMovieViewController.m
robwayne: Okay I'll follow this PR up with another PR where I do this, seeing as I have other PRs out with additional code that would need to be added as well.

Mikanbu:
Merged with 92c0b74b8c6e7817faa12a95476242cad066a2df.

PR #353, opened 2019-07-24
VLCPlayerDisplayController: Miniplayer: Check is player is setup

Mikanbu:
### Checklist

### Description

This makes the property playerIsSetup accessible in order to check if the player is setup.

Indeed for the case of the miniplayer, we were basing the shown
state on the return values of libvlc `willPlay` and `isPlaying`.
This led to have inconsistency because of synchronization issues.

(closes #472)


carolanitz:
LGTM

PR #356, opened 2019-08-06
MedialibraryService: Only request thumbnail when media is there

carolanitz:
### Checklist

### DIFF
--- a/SharedSources/MediaLibraryService.swift
+++ b/SharedSources/MediaLibraryService.swift
@@ -361,7 +361,7 @@ extension MediaLibraryService {
 extension MediaLibraryService {
     func requestThumbnail(for media: [VLCMLMedia]) {
         media.forEach() {
-            guard !$0.isThumbnailGenerated() else { return }
+            if let thumbnail = $0.thumbnail(of: .thumbnail), !thumbnail.path.isEmpty { return }
             if !$0.requestThumbnail(of: .thumbnail, desiredWidth: 320, desiredHeight: 200, atPosition: 0.3) {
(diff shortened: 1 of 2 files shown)


SharedSources/MediaLibraryService.swift
```diff
@@ -413,8 +413,11 @@ extension MediaLibraryService: VLCMediaFileDiscovererDelegate {
         } catch let error {
             assertionFailure("MediaLibraryService: VLCMediaFileDiscovererDelegate: \(error.localizedDescription)")
         }
-
-        reload()
+        if let media = fetchMedia(with: excludeURL) {
+            requestThumbnail(for: [media])
```
Mikanbu: Are you sure this works? I believe that since the media haven't been added yet, `fetchMedia` will mostly return nil.

This is one of the reason of existence of the patch I sent you privately.

SharedSources/MediaLibraryService.swift
```diff
@@ -361,7 +361,7 @@ extension MediaLibraryService {
 extension MediaLibraryService {
     func requestThumbnail(for media: [VLCMLMedia]) {
         media.forEach() {
-            guard !$0.isThumbnailGenerated() else { return }
+            if let thumbnail = $0.thumbnail(of: .thumbnail), !thumbnail.path.isEmpty { return }
```
Mikanbu: `isThumbnailGenerated` should do the trick here since we always use `.thumbnail`.

Additionally, with the next release of the wrapper, nil will be returned on empty MRLs.

PR #358, opened 2019-08-06
Remove image constraint

dnicolson:
### Checklist

### Description
This pull request removes the constraint on the image when a local network source is selected.

Before:
bad

After:
ezgif-1-779e8f6cb083

Perhaps the background should be removed entirely, as it doesn't suit the generic icon well:

generic

There is also a flicker of the image when the view first loads:

ezgif-1-ef4e5d1b20de

The Simulator would not accept the app until the CFBundleVersion was changed to only [include numbers](https://stackoverflow.com/a/56536439/2693925).

### DIFF
--- a/Apple-TV/Info.plist
+++ b/Apple-TV/Info.plist
@@ -21,8 +21,7 @@
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string>325
-</string>
+	<string>325</string>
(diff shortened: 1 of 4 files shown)


Mikanbu:
Merged with 5b09a5dbb16c9e85523cf391dbfcea6611ec1606, d837330ced9cc7f7dca8bcfc8d90b9ecdd4ecd28.

PR #379, opened 2019-08-19
(WIP) MediaMoreOptionsActionSheet: Present PlaybackSpeedView on playback selection

robwayne:
So currently, I'm trying to incorporate the `PlaybackSpeedView` into the `MediaMoreOptionsActionSheet` however, whenever I try to load the nib, in lines 87-92 there is a crash because of the not satisfying the  key-compliance of the audioDelayIndicator. This obviously doesn't happen when the PlaybackSpeedView is opened from the PlaybackSpeedView button in the `MovieViewController`. I have tried adding reconnecting the outlet for the audioDelayIndicator to no avail. 

### Checklist
- [ ] I've followed the [vlc-ios code style](Docs/CodingStyle.md)
- [ ] I've read the [Contribution Guidelines](https://github.com/videolan/vlc-ios#contribute)

### DIFF
--- a/Sources/ActionSheet/MediaMoreOptionsActionSheet.swift
+++ b/Sources/ActionSheet/MediaMoreOptionsActionSheet.swift
@@ -84,6 +84,15 @@ extension MediaMoreOptionsActionSheet: MediaPlayerActionSheetDataSource {
             if $0 == .interfaceLock {
                 cellModel.accessoryType = .toggleSwitch
                 cellModel.viewToPresent = nil
+            } else if $0 == .playback {
+                if let playbackView = Bundle.main.loadNibNamed("PlaybackSpeedView", owner: self, options: nil)?.first
+                    as? PlaybackSpeedView {
+                    cellModel.viewToPresent = playbackView
+                } else {
+                    print("MoreOptionsSheet: PlaybackSpeedView could not be loaded from Nib")
+                }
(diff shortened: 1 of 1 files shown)


Sources/ActionSheet/MediaMoreOptionsActionSheet.swift
```diff
@@ -84,6 +84,13 @@ extension MediaMoreOptionsActionSheet: MediaPlayerActionSheetDataSource {
             if $0 == .interfaceLock {
                 cellModel.accessoryType = .toggleSwitch
                 cellModel.viewToPresent = nil
+            } else if $0 == .playback {
+                if let playbackView = Bundle.main.loadNibNamed("PlaybackSpeedView", owner: self, options: nil)?.first
```
robwayne: trying to instantiate the nib with `nib.instantiate(name:owner:options)` also fails. I think this crash in general might be due to the fact that in the `PlaybackSpeedView`'s `awakeFromNib` method, a `loadNib` function is called that itself calls the `nib.instantiate(name:owner:options)` method. This can be found at line `56` in the `PlaybackSpeedView.swift` file.

fkuehne:
This was superseded by https://code.videolan.org/videolan/vlc-ios/-/merge_requests/283 that was subsequently merged.

PR #385, opened 2019-08-22
CollectionModel: Iterate over files from the back to fix wrong deletion

carolanitz:
closes #604

### Checklist

### DIFF
--- a/Sources/EditController.swift
+++ b/Sources/EditController.swift
@@ -169,7 +169,7 @@ extension EditController: EditToolbarDelegate {
     func editToolbarDidDelete(_ editToolbar: EditToolbar) {
         var objectsToDelete = [VLCMLObject]()
 
-        for indexPath in selectedCellIndexPaths {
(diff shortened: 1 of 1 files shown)


carolanitz:
> As I stated, this won't work in every case which is why ideally it would help to have a new API from the medialibrary.
> 
> Let's say we have a playlist of 5 media.
> This will probably work in a case where the user selects to delete in a precise order(i.e.: 1, 2, 3)
> 
> In a case where the users would for example select 2, 5, 1, 3 this won't remove all the selected media. The same thing would probably happen if you select 1, 2, 3, 4, 5.
> 
> So for it to work correctly without the medialibrary change, a solution would be to shift indexes after a deletion.

okay I see your point. I adjusted the change so that we add medias for most cases from the back by sorting the selectedIndexes

carolanitz:
merged with 007f39d7de592caa1bc3483b123a55ce6888e252

PR #396, opened 2019-08-27
Migrate Hockey to AppCenter

Mikanbu:
### Checklist

### DIFF
--- a/Apple-TV/AppleTVAppDelegate.m
+++ b/Apple-TV/AppleTVAppDelegate.m
@@ -16,7 +16,10 @@
 #import "VLCCloudServicesTVViewController.h"
 #import "VLCHTTPUploaderController.h"
 #import "VLCRemotePlaybackViewController.h"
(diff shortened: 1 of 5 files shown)


Mikanbu:
Small issue that we need to consider.

For tvOS, it seems that AppCenter have a minimum target of tvOS 11, and we still support tvOS 10.2.

carolanitz:
@Mikanbu can we wait with this until we have 3.2.0 out? I believe we still have 90 days and for the initial release It might be better to go with crashreporting that we're familiar with?

Mikanbu:
@carolanitz I would be for migrating this. It doesn't seem that there are a lot of changes.

jbkempf:
tvOS 10 is not an issue

Mikanbu:
Rebased to master and include the tvOS migration as well.

jbkempf:
LGTM

PR #397, opened 2019-08-27
MediaLibraryService: Handle dbReset correctly

Mikanbu:
### Checklist


carolanitz:
lgtm but a bit more what the issue was before and why it now needs to be handled this way either in the description or commit would be nice :)

Mikanbu:
Addressed review and rebased to master.

PR #398, opened 2019-08-27
VLCServerListViewController: localNetworkTableView: Fix cell sizing

Mikanbu:
### Checklist

### Description

By setting the estimatedRowHeight to 0, we disable estimated heights.
Indeed, it seems that the height was wrongly estimated by the system before.

(closes #627)


Mikanbu:
I just thought about it, but we could also estimate the height to the cell height.

carolanitz:
estimating it to the cellheight is better than 0

PR #405, opened 2019-08-30
MediaCategoryViewController: Hide mini player on Edit

Mikanbu:
### Checklist

### Description

This mainly set VLCPlayerDisplayController as a service and ensures that the miniplayer is hidden while editing.

Additionally, set brings minor cosmetic and removal of unnecessary imports.

(closes #626)


Mikanbu:
Update: Fix minor typo in commit message.

carolanitz:
I haven't looked at the details but I believe hiding is not a good user experience because you either stop playback with it, and that might not be wanted if you have a party and just want to add songs to a playlist or you loose the inability to control playback while editing. 
Both really not great so you need both controls visible.

There are a couple of solutions that were discussed with Louis but since they are complicated to implement we didn't do it for 3.2. so far

1. have the editbar overlay the tabbar
2. have a tripple tabbar where edit is sliding in above the miniplayer and 

1 would be preferred

Mikanbu:
Yes. As the `WIP` label states, this is a work in progress.

jbkempf:
Using a different direction. Closing.

PR #414, opened 2019-09-05
MediaCategoryViewController: Hide MiniPlayer on Edit

Mikanbu:
### Checklist

### Description

For now, this hide the mini player when entering edit mode.
This uses a behaviour already in place inside `VLCPlayerDisplayController`. 

Therefore, this doesn't change from the old behaviour.

This might not be ideal, even could be considered as a workaround, but I believe this is compromise we could agree on for the 3.2.0 release.

Closes #626


Mikanbu:
Changed to `setPlayerHidden`, is it slightly better?

jbkempf:
Not perfect, but good enough for now.

Mikanbu:
Rebased to master.

PR #420, opened 2019-09-13
ColorPalette: Fix statusBarStyle for iOS13 dark theme when VLC is light

edrflt:
Set iOS theme to dark and then VLC theme to light,
the status bar in VLC had white text on white background.

### Checklist

### Description

Change brightPalette to use a variable value for statusBarStyle in order to work on both iOS 13 and iOS 12 and lower.


Mikanbu:
What would you think about adding this property inside the `UIStatusBarStyle` context, for example:

```
extension UIStatusBarStyle {
    static var autoDarkContent: UIStatusBarStyle {
        if #available(iOS 13.0, *) {
            return .darkContent
        } else {
            return .default
        }
    }
}
```

This would contain it inside the `UIStatusBarStyle` so that we could use it as such:

`UIStatusBarStyle.autoDarkContent`

Nit-pick: `autoDarkColor` -> `autoDarkContent` to be somewhat in sync with Apple.

Mikanbu:
Merged with f3cd4a5.

PR #437, opened 2020-06-18
#trivial fix localized string error

AndrewSB:
Saw that genstrings didn't like this while I was working on something else, also re-ordered the build phases to have genstrings run earlier


Mikanbu:
Thanks for the fixes!

I took the liberty of separating into two commits: 3d0362a4 and a65cfbfd.

Cheers!

PR #438, opened 2020-06-18
Create preview for VLCMovieViewControlPanelView

AndrewSB:
This is a work in progress, I want to add some more spacing to the control panel between the volume slider & the forward/play/back controls


AndrewSB:
I initially re-cloned vlc to add PiP support, I'm having some trouble understanding how `vpc` renders. It looks like I pass it a UIView and it renders to that. I need an AVPlayerLayer instead, is there anywhere I can read documentation about the `vpc` rendering process?

Mikanbu:
@AndrewSB Thank you for the PR!

Just to let you know, we are currently in a process of rewriting the video player.

It should be released with the next version.

AndrewSB:
I'll update this when I finish it then. Once the new one is written, I'll take a look at adding PiP if y'all haven't thought of adding it in the rewrite!

AndrewSB:
is that new version out? is there documentation or an announcement that came out with it?

fkuehne:
The re-written player was merged in https://code.videolan.org/videolan/vlc-ios/-/merge_requests/283 but note that VLC cannot support PiP on iOS/iPadOS because it does not use AVPlayer.

PR #449, opened 2022-03-21
Fix shuffle behaviour

roryclear:
### Checklist
- [x] I've run `bundle exec fastlane test` from the `fastlane` directory to see all new and existing tests pass

### Description
Shuffling is now done when the shuffle button is pressed, making an array of unique positions in the playlist. Behaviour now matches other music apps.

### DIFF
--- a/Sources/VLCPlaybackService.m
+++ b/Sources/VLCPlaybackService.m
@@ -65,7 +65,6 @@ NSString *const VLCPlaybackServicePlaybackPositionUpdated = @"VLCPlaybackService
 
     NSLock *_playbackSessionManagementLock;
 
-    NSMutableArray *_shuffleStack;
     void (^_playbackCompletion)(BOOL success);
 
     VLCDialogProvider *_dialogProvider;
@@ -73,6 +72,9 @@ NSString *const VLCPlaybackServicePlaybackPositionUpdated = @"VLCPlaybackService
     VLCPlayerDisplayController *_playerDisplayController;
 
     NSMutableArray *_openedLocalURLs;
+    
+    NSInteger _currentIndex;
+    NSMutableArray *_shuffledOrder;
 }
 
 @end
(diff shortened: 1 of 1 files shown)


roryclear:
Hello,
Thanks for getting back to me, sorry my description is unclear. I have tried to use a shuffled list of indexes that is created when shuffle mode is turned on, instead of selecting a new random song every time next/previous are pressed.

This is the behaviour I have added:

Given a user has shuffle mode on
And they have just finished or skipped Song C
When the user presses the "previous track" button
Then Song C again

Current behaviour on master:

Given a user has shuffle mode on
And they have just finished or skipped Song C
When the user presses the "previous track" button
Then a random song plays

Let me know if you have any more questions or issues.

Mikanbu:
I understand better now, you are right, this should be fixed !

Thank you for the clarification.

I will make some testing and get back to you.

Mikanbu:
This doesn't seem to work as expected, even if the shuffle mode is activated the playback is always in the same 1 - 2 - 3 order.

If you look at the old code, the shuffling logic was used with the nextMediaIndex since it is the method that is called when the media is changed.

I believe that it should remain there. For the previous media logic, we could keep the older media index and launch it specifically.

Please let me know if I've missed something while testing.

roryclear:
After shuffle is turned on, the order will remain the same until playNewMedia is called, or shuffle is turned off and on again. 

Are you by chance testing this with 3 items? In that case 1 - 2 - 3 is very likely and you should be seeing 1 - 3 - 2 the other 50% of the time if you are toggling shuffle off and on or selecting a new item, I have just tried it. 

https://user-images.githubusercontent.com/45880510/160236068-3f92cb38-c875-408f-b839-b1f49d382c46.mov

Why not keep media indexes for all tracks? As far as I can tell, every music/streaming app I have access to is doing this, as their shuffle behaviour is the same as this. I cannot speak for all users, but I would think this is the more widely expected behaviour.

Mikanbu:
Sorry for the late reply, I've tested again and it seems to be working as expected.

I will port this to our gitlab for this to be merged, you can follow the process with the following link:

https://code.videolan.org/videolan/vlc-ios/-/merge_requests/645

PR #451, opened 2022-08-16
tvOS: NetworkStream UI: Add sorting button new to old

embeddedc:
### Checklist
- [ ] I've run `bundle exec fastlane test` from the `fastlane` directory to see all new and existing tests pass
- [ ] I've followed the [vlc-ios code style](Docs/CodingStyle.md)
- [ ] I've read the [Contribution Guidelines](https://github.com/videolan/vlc-ios#contribute)

### Description

It is annoying that new stream URLs are added at the bottom to the list, so you have to scroll down to resume a recent stream.

Can we add a setting to  change this behavior? Alternatively I added a "reverse sorting" button with this PR.


fkuehne:
Why not always show the last stream first like in the iOS interface?

embeddedc:
> Why not always show the last stream first like in the iOS interface?

Does this really work differently on iOS?  The logic seems to be the same?

Also I don't understand https://github.com/videolan/vlc-ios/blob/b7bf94f1e06c38aa8c956e32207cdb2cc107520f/Sources/VLCOpenNetworkStreamViewController.m#L262
This removes the latest not the oldest entry?

Do you prefer a PR for https://github.com/embeddedc/vlc-ios/commit/6aba00bf23ba4b2ae8f41d6f12b4b31e80e23f02?

fkuehne:
I prefer to not have a button at all and always have the latest stream at the top and the oldest at the bottom. This is the behaviour on iOS and that on tvOS, it is reverse is simply a bug.

A MR for the referenced commit would be appreciated on our own gitlab here: https://code.videolan.org/videolan/vlc-ios

Thank you!

lawchingman (approved):

lawchingman (approved):

lawchingman (approved):

lawchingman (approved):

lawchingman (approved):

PR #454, opened 2022-12-25
ios: fix: last selected subtitle not recovered

pangwa:
Add a small delay to recover the playback state.

The bug reproduce steps(install VLC on iPhone)：
1. Play a video with two or more subtitles,
2. Switch to a subtitle other than the first one
3. close the video and back to the video list
4. Open the same video again
5. The selected subtitle was the first one.

Expected behavior:
At step 5, the selected subtitle should be the one last selected.

### DIFF
--- a/Sources/Playback/Control/VLCPlaybackService.m
+++ b/Sources/Playback/Control/VLCPlaybackService.m
@@ -454,8 +454,12 @@ NSString *const VLCPlaybackServicePlaybackPositionUpdated = @"VLCPlaybackService
     VLCMLMedia *media = [_delegate mediaForPlayingMedia:_mediaPlayer.media];
 
     if (media) {
-        _mediaPlayer.currentAudioTrackIndex = (int) media.audioTrackIndex;
-        _mediaPlayer.currentVideoSubTitleIndex = (int) media.subtitleTrackIndex;
+        NSTimeInterval delayInSeconds = 0.2;
+        dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC));
+        dispatch_after(popTime, dispatch_get_main_queue(), ^(void){
+            _mediaPlayer.currentAudioTrackIndex = (int) media.audioTrackIndex;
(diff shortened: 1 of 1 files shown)


Sources/Playback/Control/VLCPlaybackService.m
```diff
@@ -454,8 +454,12 @@ - (void)restoreAudioAndSubtitleTrack
...
 
     if (media) {
-        _mediaPlayer.currentAudioTrackIndex = (int) media.audioTrackIndex;
-        _mediaPlayer.currentVideoSubTitleIndex = (int) media.subtitleTrackIndex;
+        NSTimeInterval delayInSeconds = 0.2;
+        dispatch_time_t popTime = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(delayInSeconds * NSEC_PER_SEC));
+        dispatch_after(popTime, dispatch_get_main_queue(), ^(void){
+            _mediaPlayer.currentAudioTrackIndex = (int) media.audioTrackIndex;
+            _mediaPlayer.currentVideoSubTitleIndex = 3; // (int) media.subtitleTrackIndex;
+        });
```
fkuehne, line 462: Why the dispatch?

and that constant of 3 is wrong. It can be any number.

pangwa: I believe at the time of restoreAudioAndSubtitleTrack was called, the mediaPlayer was not fully initialized. So added a short delay to ensure the mediaPlayer is fully initialized to perform other setup.

pangwa: #461 was a mistake, just fixed it.

pangwa: @fkuehne please take a look.

fkuehne:
When creating this MR, did you use MobileVLCKit 3.5.0 or 3.5.1? With the fixed event handling since the previous VLCKit version, this patch might not be necessary anymore.

pangwa:
The Podfile.lock shows I'm using MobileVLCKit 3.5.0. And the issue can be reproduced on the latest ios VLCPlayer (3.3.12 447).

pangwa:
closed due to no activity

PR #455, opened 2023-02-07
Translations: Update Turkish translation

burakdede2001:
Turkish translation: Update "Give tip" translation


fkuehne:
Thanks for the fix. This is indeed a lot better. I'll put this fix directly in our translation system, so it lands in the next update automatically!

PR #457, opened 2024-08-07
Add Mutipath TCP (MPTCP) support

Aperence:
MPTCP is a TCP extension allowing to improve network reliabilty by using mutiple network interface for the same TCP connection. Check https://www.mptcp.dev and
https://developer.apple.com/documentation/foundation/urlsessionconfiguration/improving_network_reliability_using_multipath_tcp for details.
Changes to this repository include declaring a new sharedMPTCPSession property in NSURLSession, and replacing previous uses of [NSURLSession sharedSession] to [NSURLSession sharedMPTCPSession]

### DIFF
--- a/Sources/Extensions/URLSession/NSURLSession+sharedMPTCPSession.h
+++ b/Sources/Extensions/URLSession/NSURLSession+sharedMPTCPSession.h
@@ -1,0 +1,17 @@
+/*****************************************************************************
+ * NSURLSession+sharedMPTCPSession.h
+ * VLC for iOS
+ *****************************************************************************
+ *
+ * Author: Anthony Doeraene <anthony.doeraene@hotmail.com>
+ *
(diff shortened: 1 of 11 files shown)


fkuehne:
Thanks a lot for your contribution! This is a nice addition!

Would you mind opening a merge request on our gitlab? Otherwise, I can do it for you.

Regarding your code, I believe that this shouldn't be enabled for PLEX which is limited to the local network, so we don't need the (minimal) overhead. Everywhere, it could make sense!

Aperence:
Hello,

I reverted the changes to PLEX as you recommended it
For the merge request, I don't think I have the rights yet to create one (may have I missed something ?)

Have a nice day and thank you for the fast reaction !

fkuehne:
I just approved your account request so you can do it now! :)

PR #460, opened 2025-09-14
BugFix: When shuffle is on, the first song in the album is always the first song played.

lachezar:
Ideally shuffle should start from a random song

Closes #1721


fkuehne:
This work was merged: https://code.videolan.org/videolan/vlc-ios/-/merge_requests/1792

PR #463, opened 2026-04-30
fix: enable supportedSubtitles pathway

jestress:
This PR updates the in-player Files subtitle picker to classify files using VLC’s existing supported subtitle extension list instead of only recognizing `.srt`. 

As a result, `.ass`, `.ssa`, and other already-supported subtitle formats are attached as subtitle tracks rather than being incorrectly routed as external audio.


fkuehne:
Upstreamed here: https://code.videolan.org/videolan/vlc-ios/-/merge_requests/1802

Thanks for your contribution!

